PR #10, opened 2016-11-19
PublicKey25519Proposition address validation fails

AlanVerbner:
## Description

 PublicKey25519Proposition public key address validation is failing when validating checksum.

 For example:
```scala
 val pub = PrivateKey25519Companion.generateKeys(seed)._2
 PublicKey25519Proposition.validPubKey(pub.address).isSuccess == false
```
## Solution
 While creating the address, the checksum is invoked only with pubKeyBytes but when verifiying, the AddressVersion is included (`addressBytes.dropRight(ChecksumLength)` also contains the  AddressVersion in the head after droping the checksum)


kushti:
Thanks for the submission!

PR #13, opened 2016-12-16
upgrade to scala 2.12

rtkaczyk:
This PR upgrades several libraries in order to compile Scorex for Scala 2.12

Currently the only dependencies that are blocking this are `swagger-scala-module` (it looks like [it's ready](https://github.com/swagger-api/swagger-scala-module/blob/develop/build.sbt#L13) but not yet released) and `swagger-akka-http` (which also looks ready but awaits the former).

This brings me to my proposal:
Can we extract all HTTP related code into a separate module: `scorex-http`? The reason is that code is not required for build actual live clients. It's useful for prototyping and experimenting, which is why it makes sense to make it an extra addition to `scorex-core`. This also relates to `examples` which, given this change, need to be extracted as well.

This PR removes all HTTP related code and the examples. If the proposal above is not accepted, I will try to bring them back (pending swagger artifacts release).

Other notes:
- I haven't tried cross building
- It's only been tested by running the unit tests
- Deprecations fixed
- Scala 2.12.0 was chosen (2.12.1 makes MapDB unusable, it looks like a bug in Scala compiler)


rtkaczyk:
argh, wrong repo again

Issue #23, opened 2017-04-13
Generic Boxes


mpk2:
Currently, boxes are parameterized only by proposition, with a value type of `Long`. It seems that a lot of the code is intertwined with this assumption, but there's no particular reason why box values should be limited to `Long`.

One example is using `String` as the value type for a box in order to define access restrictions to a blockchain resource.

kushti:
Could you provide some example? If it is about "access restrictions", so I assume restricting access to the block itself, then probably String field should be a part of a proposition?

mpk2:
As we are developing smart contract templates, we have chosen to store a contract shared amongst members as a box with a value type that can hold the state data for that contract (sealed by a 1-of-n proposition). It's different than the above, but it is a case where the default value of `Long` seems unnecessarily cumbersome. 

I understand that one could add additional data to the boxes, but given especially the way that boxes are handled in wallet, it seems a bit silly to do so just to end up with an extraneous Long value. 

So far, we've ended up building a set of genetic type structures that sit parallel to a lot of the existing pieces (e.g. Box, Minimal State, BoxTransaction, etc) all the way up to the NodeViewHolder and Application level, largely duplicating existing logic because of how tightly coupled they all are.

mpk2:
As far as a `String` used for access restriction, technically transactions are access-limited by access to boxes necessary for their execution. Therefore, if one wanted to create a "permission" box, one could allow elevated user permissions (i.e. the submission of particular transactions) with the presentation of a box containing a particular string value (e.g. "Admin"). This is somewhat irrelevant for the point of the discussion now, but it was where I was going with that example.

mpk2:
Any further thoughts on this @kushti ?

kushti:
@mpk2  Sorry for the long reply Matt! I imagine, a contract(a box) has some number of tokens associated with it?  Then value is needed. State could be made a part of Proposition, if transaction applicability depends on it (proposition could be e.g. a tuple of <state, set of possible transitions> then).

I never thought about token-less applications actually. I'll revise whether it it possible to do painless generalization regarding value field (another code review is coming, final before next Release Candidate version).

mpk2:
Thanks! Really enjoy the work that has been put into this project. We're building out our sections as it goes along.

[...]

PR #25, opened 2017-06-05
Pull Request for Generic Boxes Issue #23

zihehuang:
I took the liberty of converting `Box` into a generic type. I had to take `validate` out of `BoxMinimalState` since the implementation requires the `Box`'s value type to be a `Long`.

### DIFF
--- a/src/main/scala/scorex/core/transaction/box/Box.scala
+++ b/src/main/scala/scorex/core/transaction/box/Box.scala
@@ -6,14 +6,9 @@ import scorex.core.transaction.box.proposition.Proposition
 /**
   * Box is a state element locked by some proposition.
   */
-trait Box[P <: Proposition] extends BytesSerializable {
-  val value: Box.Amount
+trait Box[P <: Proposition, T] extends BytesSerializable {
+  val value: T
   val proposition: P
 
   val id: Array[Byte]
 }
-
-object Box {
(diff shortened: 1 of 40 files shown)


src/main/scala/scorex/core/transaction/box/Box.scala
```diff
@@ -6,14 +6,9 @@ import scorex.core.transaction.box.proposition.Proposition
 /**
   * Box is a state element locked by some proposition.
   */
-trait Box[P <: Proposition] extends BytesSerializable {
-  val value: Box.Amount
+trait Box[P <: Proposition, T] extends BytesSerializable {
```
ceilican: What if, instead of making `Box` more generic by changing the type of `value` from `long` to `T` and carrying `T` around as a type parameter, we made `Box` more generic simply by removing `value` entirely?   It would be up to subclasses of `Box` to have fields that contain content (e.g. values or anything else one might want to add to the box).

@kushti @catena2w @zihehuang , what do you think?

mpk2: I know this is a bit old, but that could likely work just fine based on our discussion for #23 . @zihehuang probably had the same mindset I did when we were working on this locally. It's then easy enough to create a parameterised `ValueBox` if you wanted that field, anyway.

PR #53, opened 2017-09-17
Replace assert by ensuring when appropriate

ceilican:
This PR should be merged after #52.

It is more idiomatic to use `ensuring` instead of `assert` for post-conditions.


ceilican:
Sorry for the compilation erros. (Something had gone wrong in the generation of Eclipse projects using sbt-eclipse, and this prevented me from seeing the compilation errors in the IDE.)

I have fixed the errors in my last commit. I believe this PR is ready now.

PR #67, opened 2017-10-20
Improve ReadMe for Examples

ceilican:


ceilican:
@kushti , I think this MR could be accepted too.

Issue #75, opened 2017-10-24
Node view synchronizer refactoring


kushti:
NodeViewSynchronizer is caching ids of modifiers (blocks/transactions etc) asked from network (so only objects requested are expected to be delivered):
  private val asked = mutable.Map[ModifierTypeId, mutable.Set[ModifierId]]()

However, this structure is not storing a peer being asked.

Also, we need to have another structure which contains objects delivered, in order to extract a peer when, during processing, a node is realizing that object sent is e.g. invalid (so the peer is to be banned or de-prioritized)

  private val delivered = mutable.Map[ModifierId, ConnectedPeer]()

AxiomFlow:
@kushti , maybe this issue can be closed?

PR #76, opened 2017-10-25
Keep track of peers that have been asked and peers that delivered

ceilican:
Addresses #75 .

* [x] Write tests
  - [x] Create `NodeViewSynchronizerTests` in `scorex.testkit.properties` package
  - [x] Create `NodeViewSynchronizerSpec` in `hybrid` package in `examples` submodule. 
  - [x] Clean tests
* [x] Disconnect peers that have been banned
* [x] Create a class to encapsulate operations on `asked`
* [ ] Add delivery timeout to settings
* [x] Penalize peer that doesn't deliver within timeout (currently: the penalty is a total ban)
* [x] Generalize the `Asked` class and use it for `delivered` as well


ceilican:
## Banning

Relevant for banning: https://www.usenix.org/system/files/conference/usenixsecurity15/sec15-paper-heilman.pdf

Layers:   NetworkController <-> NodeViewSynchronizer <-> NodeViewHolder

to ban a peer, we need to send a message to NetworkController to drop a connection (NetworkController will send PeerConnectionHandler.CloseConnection to a peer of interest)

ceilican:
## Testing

Relevant paper: https://github.com/kushti/scorex-laws

kushti:
Please do TTL to be set via settings for asked. If something is asked but not delivered within timeout, a peer should get a negative feedback.

ceilican:
> If something is asked but not delivered within timeout, a peer should get a negative feedback.

I see two ways of doing that. The most straighforward way would be to: make NVS send a scheduled message to itself, with the given timeout. And when NVS receives the message from itself, it checks whether the asked object has been delivered.

A potential problem with this solution is that self-message would go into the queue of messages received by NVS, and there is no guarantee that it will be processed exactly after the timeout. If there are too many messages in the queue, NVS will only process the self-message and check the delivery after all previously queued messages have been processed. And this might happen only some time after the timeout.  But maybe this problem is not an issue in practice? Should I worry about this problem?

The other way would be to create a Future for the task of checking the delivery and executing the Future after the timeout. This approach would avoid the problem described above, but it would b eless elegant, I think. 

I am in favour of the self-message approach. What do you think?

kushti:
A more simple option, we can check "asked" collection on a message delivery, so along with the check that delivered object is in "asked", also filter "asked"

ceilican:
> A more simple option, we can check "asked" collection on a message delivery, so along with the check that delivered object is in "asked", also filter "asked"

Unless I misunderstood your suggestion, I think it doesn't always solve the problem and, when it does, it doesn't do it as efficiently as possible.

**Why I think it doesn't always solve the problem:** Imagine the following extreme case: if messages are never delivered, we will never check the "asked" collection, and we might never notice that there are elements in "asked" that have already exceeded the timeout.    (Less extremely, if messages are delivered on average every X milliseconds, where X is greater than the specified timeout, we would be checking "asked" not as frequently as we should be checking it...)  This wouldn't be a problem in practice, if messages are delivered frequently enough.

**Why I think it is not so efficient:** if we have N elements in "asked", we would have to traverse, check and filter N elements after every delivery. In practice, I believe we expect to have as many deliveries as asked messages. Therefore, this approach would have, roughly, quadratic complexity (Omega(N^2)).  
In contrast, The two solutions that I described above could be implemented in O(N), I believe.

 Does this make sense? Or have I misunderstood your suggestion?

ceilican:
@kushti , while testing NVS, I noticed something that seems to be a bug. I fixed it in this commit: https://github.com/ScorexFoundation/Scorex/pull/76/commits/6f855a3f5a6ea6a3100cef6bf53c278995b59b70

Without my fix, `extOpt.get` throws an exception when `extOpt` is empty.

ceilican:
@kushti , this MR is ready to be reviewed and, hopefully, accepted.

For now:

1. The delivery timeout is hard-coded. (I will keep trying to understand why Ficus is not managing to parse the config file when I add "deliveryTimeout" to "settings.conf" and to the "NetworkSettings" case class.)

2. Peers that do not deliver on time are *banned*.  I believe a nicer de-prioritaziation scheme (instead of total banning) would require more extensive changes in PeerInfo and PeerManager. I sketched a proposal here: https://github.com/ScorexFoundation/Scorex/pull/76/commits/6c3f90392e126a0cf78567bb98b4d4f8529d641b . Since this goes beyond refactoring NodeViewSynchronizer, I think we should work on this in another MR, after this MR is accepted.

ceilican:
Thanks!

PR #103, opened 2017-11-03
Integrate NVHTests and NVSTests in BlockchainSanity

### LINKED ISSUE #101: Plug NodeViewHolderTests / NodeViewSynchronizerTests into BlockchainSanity

ceilican:
Closes #101.

* [x] Fix incompatibility between `org.scalatest.PropSpec` and `org.scalatest.fixture.PropSpec`.  (NVHTests and NVSTests extend `fixture.PropSpec`, whereas other test traits mixed in BlockchainSanity extend `PropSpec`. Therefore, they currently cannot be combined together.)


ceilican:
```
[info] Compiling 2 Scala sources to /Users/Bruno/Dropbox/Code/Scorex/testkit/target/scala-2.12/classes ...
[error] /Users/Bruno/Dropbox/Code/Scorex/testkit/src/main/scala/scorex/testkit/BlockchainSanity.scala:39:10: illegal inheritance; superclass PropSpec
[error]  is not a subclass of the superclass PropSpec
[error]  of the mixin trait NodeViewHolderTests
[error]     with NodeViewHolderTests[P, TX, PM, ST, SI, HT, MPool]
[error]          ^
[error] /Users/Bruno/Dropbox/Code/Scorex/testkit/src/main/scala/scorex/testkit/BlockchainSanity.scala:40:10: illegal inheritance; superclass PropSpec
[error]  is not a subclass of the superclass PropSpec
[error]  of the mixin trait NodeViewSynchronizerTests
[error]     with NodeViewSynchronizerTests[P, TX, PM, ST, SI, HT] {
[error]          ^
[error] two errors found
[error] (testkit/compile:compileIncremental) Compilation failed
```

ceilican:
There might be different ways to solve this problem:

1. Make other test traits extend `fixture.PropSpec` instead of `PropSpec`.
2. Make NVHTests and NVSTests extend `PropSpec` instead of `fixture.PropSpec`.
3. Have two separate `BlockchainSanity` traits, one for test traits extending `fixture.PropSpec` and another for test traits extending `PropSpec`.

@kushti , do you have any preference?

kushti:
@ceilican lets make other test traits extend fixture.PropSpec instead of PropSpec ?

ceilican:
Ok.

ceilican:
@kushti , after reading the relevant parts of the ScalaTest documentation (http://www.scalatest.org/user_guide/sharing_fixtures), I believe the best solution would be option 2 described above. When I wrote the 3 options, I was unsure about how to deal with fixtures if NVHTests and NVSTests inherited `PropSpec` instead of `fixture.PropSpec`, but now I see that we could use the loan-fixture method pattern described here: http://www.scalatest.org/user_guide/sharing_fixtures . 

Advantages of option 2 would include:

1. We wouldn't need to implement dummy fixtures for tests that actually do not need fixtures. (We would have to do that with option 1.)
2. Only NVSTests and NVHTests would have to be refactored. (Option 1 would require refactoring all other tests.)
3. We would be able to share loan-fixture methods between NVSTests and NVHTests, thereby reducing code duplication slightly.
4. NVSTests and NVHTests would become more elegant, because we would have more control over which fixture objects need to be passed to each test. (Right now, we are passing a big tuple of objects as fixture to all tests, even though most tests need only some elements of that tuple. Passing only objects that are really needed by each test would make test execution more efficient.)
5. We would conform with the recommendation in [fixture.PropSpec's documentation](http://doc.scalatest.org/3.0.1/#org.scalatest.fixture.PropSpec), which says that "fixture.PropSpec is intended for use in special situations, with class PropSpec used for general needs."

May I implement option 2 instead of option 1?

kushti:
Ok, let's stick with the second option

ceilican:
@kushti , it is done. It can be accepted now.

PR #121, opened 2017-12-05
Sync rework: Part I

kushti:


src/main/scala/scorex/core/consensus/History.scala
```diff
@@ -104,6 +104,7 @@ object History {
     val Younger = Value(2)
     val Older = Value(3)
     val Nonsense = Value(4)
```
catena2w: Add comments, what is the correct response for what king of situation. Where should we return Unknown, where should we return Nonsense?

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -140,6 +140,8 @@ class NetworkController(settings: NetworkSettings,
       peer.handlerRef ! PeerConnectionHandler.CloseConnection
       peerManagerRef ! PeerManager.Disconnected(peer.socketAddress)
 
+    case Handshaked(remote, handshake) =>
```
catena2w: Do nothing here? Why do we send this message then?

testkit/src/main/scala/scorex/testkit/properties/NodeViewSynchronizerTests.scala
```diff
@@ -92,17 +92,6 @@ trait NodeViewSynchronizerTests[P <: Proposition,
     vhProbe.fishForMessage(3 seconds) { case m => m == GetSyncInfo }
   }
 
-  property("NodeViewSynchronizer: CurrentSyncInfo") { ctx =>
```
catena2w: Add test for current logic instead of this one?

src/main/scala/scorex/core/consensus/History.scala
kushti: Unknown is initial state, Nonsense is when we think other peer is going nuts. Not sure, maybe some additional types are needed.

src/main/scala/scorex/core/network/NetworkController.scala
kushti: removed

testkit/src/main/scala/scorex/testkit/properties/NodeViewSynchronizerTests.scala
kushti: the test checked behavior which was removed for being problematic. New tests will be later, they're non-trivial

PR #125, opened 2017-12-06
Sync rework: Part II

kushti:
Separation of syncing and stable regimes


src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -39,22 +40,24 @@ class NodeViewSynchronizer[P <: Proposition, TX <: Transaction[P], SI <: SyncInf
...
-  /*
+  /**
     We cache peers along with their statuses (whether another peer is ahead or behind of ours,
     or comparison is not possible, or status is not yet known)
-   */
+   **/
   protected val statuses = mutable.Map[ConnectedPeer, HistoryComparisonResult.Value]()
   protected val statusUpdated = mutable.Map[ConnectedPeer, Timestamp]()
-  private var lastSyncInfoSentTime = NetworkTime.time()
+
```
ceilican: Maybe we could encapsulate all this mutable state (e.g. `statuses`, `statusUpdated`) into a class `StatusTracker` (similarly to what we did for `DeliveryTracker`) and then `stableSyncRegime` could be a public `def`, instead of a `var` inside this new class.

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -39,22 +40,24 @@ class NodeViewSynchronizer[P <: Proposition, TX <: Transaction[P], SI <: SyncInf
...
+  protected var stableSyncRegime = false
+  protected var scheduler: Cancellable = _
```
ceilican: I think this is bug-prone, because different parts of the code may re-assign `scheduler` and, therefore, other parts of the code may end up canceling the wrong scheduler. If possible, I think it would be worth refactoring this in a way that protects ourselves from future bugs of this kind.

ceilican: For now I see that the only time the scheduler is re-assigned is immediately after the scheduler is canceled, in lines 195--196. So, there is no bug of the kind described above. But my concern is about potential future bugs.

ceilican:
Just curious: what practical need prompted you to implement different refresh timeout and interval for stable and non-stable sync regimes?  Was there any performance problem when the same value forany regime was being used?

ceilican:
ToDo:

* [ ] `updateLastSyncSentTime` is apparently called with different intended purposes in `syncSend` and in `peerManagementEvents`. More generally, it seems that we are not being careful enough about keeping timestamps of our interactions with peers. I think there might be bugs. See this: https://github.com/ScorexFoundation/Scorex/pull/125/commits/c5fad00bbc541134bb5da83b6319f6df6e627d1d .

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -83,70 +87,68 @@ class PeerManager(settings: ScorexSettings, timeProvider: NetworkTimeProvider) e
...
+      declaredAddress.contains(settings.network.bindAddress) ||
+      settings.network.declaredAddress.contains(address)
   }
 
+  private var lastIdUsed = 0
 
   private def peerCycle: Receive = {
-    case Connected(newPeer@ConnectedPeer(remote, _)) =>
-      if (peerDatabase.isBlacklisted(newPeer.socketAddress)) {
+    case Connecting(remote, direction) =>
```
ceilican: A minor style remark: a message with name ending with "-ing" seems strange. Maybe we should try to stick to past participle and imperative names (e.g. "DoSomething", telling an actor to do something, and "Done", informing an actor that something was done.)

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -83,70 +87,68 @@ class PeerManager(settings: ScorexSettings, timeProvider: NetworkTimeProvider) e
...
-        if (connectingPeer.contains(remote)) {
-          log.info(s"Connected to $remote. ${connectedPeers.size} connections are open")
-          connectingPeer = None
+        val refuse =
+          if (direction == Incoming) false
+          else if (connectingPeers.contains(remote)) {
+            log.info(s"Connecting to $remote")
+            false
+          } else {
+            log.info(s"Already connected peer $remote trying to connect, going to disconnect it")
```
ceilican: Are we really going to disconnect it? Or just refuse the second connection? I think this should be clarified.

src/main/scala/scorex/core/network/peer/PeerManager.scala
kushti: Fixed, it is DoConnecting now

src/main/scala/scorex/core/network/peer/PeerManager.scala
kushti: done, new message is "Already connected peer $remote trying to connect, going to drop the duplicate connection" (so original connection is still alive)

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -83,70 +87,68 @@ class PeerManager(settings: ScorexSettings, timeProvider: NetworkTimeProvider) e
...
+
+      if (peerDatabase.isBlacklisted(remote)) {
         log.info(s"Got incoming connection from blacklisted $remote")
       } else {
-        connectedPeers += newPeer -> None
-        if (connectingPeer.contains(remote)) {
-          log.info(s"Connected to $remote. ${connectedPeers.size} connections are open")
-          connectingPeer = None
+        val refuse =
+          if (direction == Incoming) false
```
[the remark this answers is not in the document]
kushti: Issue exists https://github.com/ScorexFoundation/Scorex/issues/153

src/main/resources/reference.conf
```diff
@@ -123,27 +123,22 @@ scorex {
...
     syncStatusRefresh = 2m
 
+    # Synchronization status update interval for stable regime
+    syncStatusRefreshStable = 4m
+
     # Network controller timeout
     controllerTimeout = 5s
   }
 
-  miner {
```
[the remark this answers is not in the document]
kushti: Fixed

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -45,39 +47,94 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
...
-  protected val statuses = mutable.Map[ConnectedPeer, HistoryComparisonResult.Value]()
-  /**
-    * Last time our node sent Sync messages to specified peer
+    * SyncTracker caches the peers' statuses (i.e. whether they are ahead or behind this node)
     */
-  protected val lastSyncSend = mutable.Map[ConnectedPeer, Timestamp]()
-  protected var lastSyncInfoSentTime: Long = 0L
-  protected var historyReaderOpt: Option[HR] = None
-  protected var mempoolReaderOpt: Option[MR] = None
+  object SyncTracker {
```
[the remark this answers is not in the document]
kushti: done

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -210,35 +258,27 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
...
-            val ext = extOpt.get
-            ext.groupBy(_._1).mapValues(_.map(_._2)).foreach {
-              case (mid, mods) =>
-                networkControllerRef ! SendToNetwork(Message(invSpec, Right(mid -> mods), None), SendToPeer(remote))
-            }
-          }
-        //todo: should we ban peer if its status is unknown after getting info from it?
-        case Unknown => log.warn("Peer status is still unknown")
+        case Unknown => log.warn("Peer status is still unknown") //todo: should we ban peer if its status is unknown after getting info from it?
+        case Nonsense => log.warn("Got nonsense") //todo: we should ban peer if its view is totally different from ours
```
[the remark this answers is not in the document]
kushti: Probably we should ban, but further investigation is needed, made the issue https://github.com/ScorexFoundation/Scorex/issues/158

Issue #96, opened 2017-10-31
PeerSynchronizer has a hard-coded timeout


ceilican:
Maybe it would be more elegant to obtain this timeout from a config file?

PR #131, opened 2017-12-12
Hard coded timeouts have been moved to config

daron666:


daron666:
Resolve #96

src/main/scala/scorex/core/settings/Settings.scala
```diff
@@ -42,7 +42,9 @@ case class NetworkSettings(nodeName: String,
                            maxPacketLen: Int,
                            maxInvObjects: Int,
                            syncInterval: FiniteDuration,
-                           syncStatusRefresh: FiniteDuration
+                           syncTimeout: Option[FiniteDuration],
```
ceilican: @kushti, @catena2w @daron666, maybe we could consider making `syncTimeout` have type `Option[Timeout]` instead of `Option[FiniteDuration]`...  But, anyway, if we decide to do that, I think it would be better to do it in another PR, where we could thoroughly revise the types of all settings and choose types that are as specific as possible. 

I think this PR is ready to be accepted.

daron666: I prefer FiniteDuration over Timeout, cause it's from Scala meanwhile Timeout are came from Akka library.

ceilican: Yes. That makes sense. I have just realized that akka just needs to encapsulate durations in Timeout because `?` needs an implicit argument.

PR #132, opened 2017-12-12
Duplicated code has been removed.

### LINKED ISSUE #89: Code duplication between generators in testkit and in core's test
The following two files contain repeated code:

* https://github.com/ScorexFoundation/Scorex/blob/master/testkit/src/main/scala/scorex/testkit/generators/CoreGenerators.scala

* https://github.com/ScorexFoundation/Scorex/blob/master/src/test/scala/scorex/ObjectGenerators.scala

daron666:
Resolve #89

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -63,7 +63,7 @@ val testingDependencies = Seq(
   "com.typesafe.akka" %% "akka-testkit" % "2.5.3" % "test",
   "org.scalactic" %% "scalactic" % "3.0.3" % "test",
   "org.scalatest" %% "scalatest" % "3.0.3" % "test",
-  "org.scalacheck" %% "scalacheck" % "1.13.+" % "test",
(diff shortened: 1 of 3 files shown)


build.sbt
```diff
@@ -63,7 +63,7 @@ val testingDependencies = Seq(
   "com.typesafe.akka" %% "akka-testkit" % "2.5.3" % "test",
   "org.scalactic" %% "scalactic" % "3.0.3" % "test",
   "org.scalatest" %% "scalatest" % "3.0.3" % "test",
-  "org.scalacheck" %% "scalacheck" % "1.13.+" % "test",
+  "org.scalacheck" %% "scalacheck" % "1.13.+",
```
kushti: Why "test" scope removed?

daron666: Cause to inherit this class from other module we need him to be not in tests. So that's why we need generators in main source.

PR #127, opened 2017-12-09
Remove Node Nonce

ceilican:
* [x] remove `nonce` from settings
* [x] make `bindAddress` and `declaredAddress` parse as `InetSocketAddress` instead of `String`.
* [X] investigate why one test is failing:

```
[info] - NodeViewHolder: simple forking *** FAILED ***
[info]   java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class scorex.core.NodeViewHolder$SyntacticallySuccessfulModifier
[info]   at scala.Predef$.assert(Predef.scala:219)
[info]   at akka.testkit.TestKitBase.expectMsgClass_internal(TestKit.scala:472)
[info]   at akka.testkit.TestKitBase.expectMsgType(TestKit.scala:445)
[info]   at akka.testkit.TestKitBase.expectMsgType$(TestKit.scala:445)
[info]   at akka.testkit.TestKit.expectMsgType(TestKit.scala:814)
[info]   at scorex.testkit.properties.NodeViewHolderTests.$anonfun$$init$$32(NodeViewHolderTests.scala:199)
[info]   at scorex.testkit.properties.NodeViewHolderTests.$anonfun$$init$$32$adapted(NodeViewHolderTests.scala:173)
[info]   at scorex.testkit.properties.NodeViewHolderTests.withFixture(NodeViewHolderTests.scala:47)
[info]   at scorex.testkit.properties.NodeViewHolderTests.$anonfun$$init$$31(NodeViewHolderTests.scala:173)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
```

PR #136, opened 2017-12-15
Improve api settings

ceilican:


ceilican:
This PR should be reviewed only after #127.

ceilican:
Partially addresses #134 .

PR #140, opened 2017-12-20
Swagger api

catena2w:


examples/src/main/resources/api/testApi.yaml
```diff
@@ -0,0 +1,198 @@
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/Peer'
+
+  /peers/connect:
+    post:
+      summary: Get current size of the unconfirmed transactions pool
```
AxiomFlow: Is this summary correct?

lock.sbt
```diff
@@ -5,46 +5,31 @@ dependencyOverrides in ThisBuild ++= Seq(
...
-  "com.fasterxml.jackson.core" % "jackson-annotations" % "2.8.9",
-  "com.fasterxml.jackson.core" % "jackson-core" % "2.8.9",
-  "com.fasterxml.jackson.core" % "jackson-databind" % "2.8.9",
-  "com.fasterxml.jackson.dataformat" % "jackson-dataformat-yaml" % "2.8.9",
-  "com.fasterxml.jackson.module" % "jackson-module-paranamer" % "2.8.9",
-  "com.fasterxml.jackson.module" % "jackson-module-scala_2.12" % "2.8.9",
   "com.github.mpilquist" % "simulacrum_2.12" % "0.10.0",
-  "com.github.swagger-akka-http" % "swagger-akka-http_2.12" % "0.10.0",
-  "com.google.guava" % "guava" % "20.0",
+  "com.google.guava" % "guava" % "19.0",
```
AxiomFlow: Why go down from 20.0 to 19.0?

src/main/scala/scorex/core/api/http/ApiRoute.scala
```diff
@@ -15,25 +15,32 @@ trait ApiRoute extends Directives {
   val context: ActorRefFactory
   val route: Route
 
-  implicit val timeout = Timeout(settings.timeout)
+  implicit lazy val timeout: Timeout = Timeout(settings.timeout)
```
AxiomFlow: Is this `timeout` still used? Maybe this line could be removed?

AxiomFlow: Or is it required as an implicit argument somewhere?

ceilican:
@Tolsi , the CI build is failing.

examples/src/main/resources/api/testApi.yaml
```diff
@@ -0,0 +1,193 @@
...
```
ceilican: Is this the correct summary for this path?

daron666: @ceilican I think not. I fixed this one here https://github.com/ergoplatform/ergo/pull/108/files#diff-f11bc37099101a6156775fbde1871902R712

ceilican: Your fix seems good to me.

daron666: @ceilican But it was done in ergo repository =)

ceilican: Yes, I know. I think it should be done here (in Scorex, in this PR) too.

And I wonder if it would be possible/desirable, in the future (in another PR), to reduce code and documentation duplication at the api level between scorex('s example) and ergo.  It seems to me that right now duplication is only avoided at the network layer. All other layers remain completely abstract in Scorex's core and need to be (re-)implemented, possibly in similar ways, by every blockchain system built with Scorex. Perhaps, in the future, Scorex could provide some standard api routes/paths already pre-implemented.

@daron666, is ergo's api significantly different from the api of scorex's example?

@kushti, should we include the api into the scope of that architecture review that we have discussed?

daron666: @ceilican Ergo's api is a superset over scorex api.

catena2w:
@ceilican review pleasse

lock.sbt
catena2w: Library, that required guava 20 was removed, we now only have libraries that require guava 19 in dependencies

src/main/scala/scorex/core/api/http/ApiRoute.scala
catena2w: It is required in a lot of API routes as a timeout for requests to actors. Check `ApiRouteWithFullView` for example

PR #143, opened 2017-12-28
Ntp config

Tolsi:


examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
```diff
@@ -129,7 +130,7 @@ class HybridHistory(val storage: HistoryStorage,
       }
     }
     // require(modifications.toApply.exists(_.id sameElements powBlock.id))
-    (new HybridHistory(storage, settings, validators, statsLogger), progress)
+    (new HybridHistory(storage, settings, validators, statsLogger, timePrivider), progress)
```
ceilican: "timePrivider" --> "timeProvider"

src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -9,31 +9,28 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
 
-  private type State = Either[(NetworkTime, Future[NetworkTime]), NetworkTime]
+protected case class NetworkTime(offset: NetworkTime.Offset, lastUpdate: NetworkTime.Time)
+
+case class NtpTimeProviderSettings(server: String, updateEvery: FiniteDuration, timeout: FiniteDuration)
 
-  private val TimeTillUpdate = 1000 * 60 * 30L
-  // 30 minutes
-  private val NtpServer = "pool.ntp.org"
+class NtpTimeProvider(ntpSettings: NtpTimeProviderSettings) extends ScorexLogging {
```
ceilican: What does "Ntp" mean? Does it mean "NetworkTimeProvider"? If so, I would suggest renaming "NtpTimeProvider" to "NetworkTimeProvider", to avoid redundancy in the name.

src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -9,31 +9,28 @@ import scala.concurrent.ExecutionContext.Implicits.global
...
-
-object NetworkTime extends ScorexLogging {
   type Offset = Long
   type Time = Long
+}
 
-  private type State = Either[(NetworkTime, Future[NetworkTime]), NetworkTime]
+protected case class NetworkTime(offset: NetworkTime.Offset, lastUpdate: NetworkTime.Time)
+
+case class NtpTimeProviderSettings(server: String, updateEvery: FiniteDuration, timeout: FiniteDuration)
```
ceilican: What does "Ntp" mean? Does it mean "NetworkTimeProvider"? If so, I would suggest renaming "NtpTimeProviderSettings" to "NetworkTimeProviderSettings", to avoid redundancy in the name.

ceilican: I think this class should be moved to "settings.scala", together with all the other settings.

PR #156, opened 2018-01-26
Add `PeerDatabaseImpl` tests

greenhat:
I took a liberty and cover `PeerDatabaseImpl` with test decoupling some dependencies along the way. It's ready to be reviewed. Your feedback is appreciated.


greenhat:
I see changes to `PeerInfo` in #155. I'll rebase when  #155 will be merged.

greenhat:
Please note that I did not cover the "self exclusion" logic in `knownPeers`. I'd like to suggest to make `PeerManager` responsible for not letting "self" get into the `PeerDatabase` and thus removing all the checks for "self" in `PeerDatabase`. The reasoning for this is the following:
1.  `PeerDatabase.knownPeers` calls with explicit "self inclusion" are used only in `PeerManager`'s `KnownPeers` and `RandomPeers` but not in the `RandomPeer`. 
2. Quite an extensive check if peer address is "self" is already present in `PeerManager.isSelf` and  used is handshake process to prevent "self" to be added to `PeerDatabase`;
3. Knowing about "self" peer seems like too much for a generic peer DB.

Let me know what you think. I'm ready to tackle this issue.

daron666:
@kushti @ceilican what do you think? I agree. Database responsible for storing data, and service should take care about no saving self into it.

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
```diff
@@ -10,13 +10,13 @@ import scala.collection.JavaConverters._
 
 
 //todo: persistence
-class PeerDatabaseImpl(settings: NetworkSettings,
-                       filename: Option[String],
-                       timeProvider: NetworkTimeProvider) extends PeerDatabase {
+class PeerDatabaseImpl(bindAddress: InetSocketAddress,
```
ceilican: I think this is a good change.

@kushti, in general, I think we should be passing only parameters that are really needed for each component, instead of passing a big Settings object containing all possible parameters around everywhere.  What do you think? Should we add a task like this to the architecture review?

ceilican: A disadvantage of doing that, though, is that the parameter list becomes longer. Before we had only `settings: NetworkSettings`; now we have `bindAddress: InetSocketAddress` and `declaredAddress: Option[InetSocketAddress]`. Not a big deal in this case, but could be inconvenient when the component needs several parameters.

A middle-ground solution would be to split `NetworkSettings` into smaller settings. For instance, `bindAddress` and `declaredAddress` could be part of a `SelfSettings` inside `NetworkSettings`, and then we would pass `SelfSettings` to `PeerDatabaseImpl`.

ceilican: Also, anyway, if we move the self detection logic outside `PeerDatabaseImpl`, then we wouldn't even need to pass `bindAddress` and `declaredAddress` to `PeerDatabaseImpl`.

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
```diff
@@ -10,13 +10,13 @@ import scala.collection.JavaConverters._
...
+                       declaredAddress: Option[InetSocketAddress],
+                       filename: Option[String]) extends PeerDatabase {
```
ceilican: Do we really want to use a global `TimeProvider` instead of passing locally as an argument?

ceilican:
I agree with the proposal of moving the self-exclusion logic.

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
```diff
@@ -27,9 +27,9 @@ class PeerDatabaseImpl(settings: NetworkSettings,
     whitelistPersistence.put(address, updatedPeerInfo)
   }
 
-  override def blacklistPeer(address: InetSocketAddress): Unit = {
+  override def blacklistPeer(address: InetSocketAddress, time: NetworkTime.Time): Unit = {
```
greenhat: @ceilican We don't use the global `TimeProvider`. We explicitly pass the timestamp itself in `blacklistPeer`. So `PeerDatabaseImpl` does not know anything about `TimeProvider`.

greenhat:
@ceilican @daron666 Thanks for the review. Glad you liked the idea of extraction of the "self-exclusion" logic.
I missed that #155 was against `sync-rework` branch and not the `master`. I guess it does not make sense to cherry pick commits to my branch. 
@kushti I can close this PR and make a new one from your branch if it makes the review and merging easier.

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
ceilican: Oh, yes. I missed that. I had just seen that you were not passing NetworkTimeProvider as an argument to the constructor of `PeerDatabaseImpl` anymore, and then I wrongly assumed that you would be accessing `NetworkTime` globally.

Now I see better what you did, and I think it was a good change. I think it is a good idea to pass `time` explicitly here, and let `PeerManager`, instead of `PeerDatabaseImpl`, worry about time and the network time provider.

greenhat:
PR is up to date with `master`. I've created #161 for my next task of "self-exclusion" logic extraction.

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
kushti: Reasonable change

src/main/scala/scorex/core/network/peer/PeerDatabaseImpl.scala
kushti: @ceilican finer granularity for settings sounds like a good choice. If more than 2 parameters to pass, it is better use *Settings I suppose. Two params are ok.

kushti:
@greenhat going to merge, thanks a lot for the contributions!

PR #160, opened 2018-01-30
NodeViewSynchronizer refactoring

kushti:


kushti:
@ceilican @catena2w some refactoring + clearing peer status on disconnect

PR #164, opened 2018-01-31
added cors allowed origin param to rest api section. also base apirou…

### LINKED ISSUE #27: Make Access Control Allowed Origin tunable
Currently, all the `settings.json` files have a field called `rpcAllowed`, but it is not used anywhere. I would assume that this was designed to take in some origins for which CORS would be allowed. 

At the same time, when you set `cors` to be `true`. `Access-Control-Allow-Origin` will be set to `*` in `ApiRoute.scala`.

Wouldn't it make more sense to control the `Access-Control-Allow-Origin` with the origins specified in `rpcAllowed`?

daron666:
…te has been reworked.

fixes #27


examples/src/main/scala/examples/commons/SimpleBoxTransaction.scala
```diff
@@ -61,7 +61,7 @@ case class SimpleBoxTransaction(from: IndexedSeq[(PublicKey25519Proposition, Non
 
   override lazy val json: Json = Map(
     "id" -> Base58.encode(id).asJson,
-    "newBoxes" -> newBoxes.map(b => Base58.encode(b.id).asJson).asJson,
+    "newBoxes" -> newBoxes.map(b => Base58.encode(b.id).asJson).toSeq.asJson,
```
ceilican: Why is this change needed?

src/main/resources/reference.conf
```diff
@@ -27,6 +27,9 @@ scorex {
     # Enable/disable CORS support
     corsAllowed = false
 
+    # Allowed origin line for cors header value
+    corsAllowedOrigin = "*"
```
ceilican: Instead of having `corsAllowed: Boolean` and `corsAllowedOrigin: String`, we could have just a single `corsAllowedOrigin: Option[String]`.  When it is `None`, cors is not allowed. When it is `Some(s)`, cors is allowed with origin `s`.

src/main/scala/scorex/core/api/http/ActorHelper.scala
```diff
@@ -0,0 +1,15 @@
+package scorex.core.api.http
+
+import akka.pattern.ask
+import akka.actor.ActorRef
+import akka.util.Timeout
+
+import scala.concurrent.Future
+import scala.reflect.ClassTag
+
+trait ActorHelper {
```
ceilican: This trait is not specific to this package and could be useful elsewhere in scorex. I would suggest moving it to `scorex.core.utils`.

examples/src/main/scala/examples/commons/SimpleBoxTransaction.scala
daron666: I don't know why, but after updating circe's version this started to produce compile error.

src/main/scala/scorex/core/api/http/ActorHelper.scala
daron666: I think so too. So I will work according your comments and prepare pr in the nearest future.

daron666:
created #170

PR #166, opened 2018-02-02
Refactoring in syncing code

kushti:


src/main/scala/scorex/core/network/SyncTracker.scala
```diff
@@ -65,12 +65,20 @@ class SyncTracker(nvsRef: ActorRef,
     }
   }
 
-  def clearStatus(remote: InetSocketAddress): Unit =
+
+  //todo: combine both?
```
ceilican: both what?

kushti: Both lines starting with "statuses.find" and "lastSyncSentTime.find" (logic is the same: look for a peer with the given socket address and remove it from a map if found

Issue #172, opened 2018-02-07
Add scalastyle report


daron666:

EzequielPostan:
Hi @daron666,

I can work on this one if you want help or someone to free your time.

daron666:
@EzequielPostan Sorry, I already finished this one =)

EzequielPostan:
hahaha, no worries  @daron666 :P

EzequielPostan:
Hi @daron666, should this issue be closed?

daron666:
@EzequielPostan I think so.

PR #174, opened 2018-02-08
fix scalastyle errors and warnings

daron666:


daron666:
fixes #172

src/main/scala/scorex/core/api/http/NodeViewApiRoute.scala
```diff
@@ -36,7 +36,7 @@ case class NodeViewApiRoute[P <: Proposition, TX <: Transaction[P]]
   type VL <: Vault[P, TX, PM, _ <: Vault[P, TX, PM, _]]
 
   //TODO null?
-  private val source: ConnectedPeer = null
+  //private val source: ConnectedPeer = null
```
ceilican: If `source` is not needed, maybe we shoudl removed this line completely, instead of just commenting it out?

src/main/scala/scorex/core/consensus/History.scala
```diff
@@ -5,7 +5,7 @@ import scorex.core.consensus.History.ProgressInfo
 import scorex.crypto.encode.Base58
 
 import scala.util.Try
-
+// scalastyle:off
```
ceilican: Why do we need to switch it off here? What was it complaining about in the comment? Maybe we should switch only the triggered rule, instead of switching off all rules?

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -97,19 +97,23 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
     networkControllerRef ! SendToNetwork(msg, Broadcast)
   }
 
-  protected def viewHolderEvents: Receive = {
+  private def transactionEvents: Receive = {
```
ceilican: This change seems ok to me. But if we intend to allow users of Scorex to create blockchain systems that subclass `NodeViewSynchronizer` and override its `Receive` methods, then we should keep this `protected` instead of `private`.

@kushti, what do you think?

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -121,6 +125,8 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
       mempoolReaderOpt = Some(reader)
   }
 
+  protected def viewHolderEvents: Receive = transactionEvents orElse modifierEvents orElse historyMemPoolChanges
```
ceilican: I think we should keep them all `private` or all `protected`.

src/main/scala/scorex/core/network/PeerConnectionHandler.scala
```diff
@@ -39,7 +39,7 @@ case class ConnectedPeer(socketAddress: InetSocketAddress,
   override def equals(obj: Any): Boolean =
     obj.cast[ConnectedPeer].exists(p => p.socketAddress == this.socketAddress && p.direction == this.direction)
 
-  override def toString = s"ConnectedPeer($socketAddress)"
+  override def toString: String = s"ConnectedPeer($socketAddress)"
```
ceilican: Recently, `direction` and `handshake` were added to `ConnectedPeer`. I wonder if we should update the `toString` method to show these new fields as well.

ceilican: @kushti, what do you think?

src/main/scala/scorex/core/api/http/NodeViewApiRoute.scala
daron666: I think so

src/main/scala/scorex/core/consensus/History.scala
daron666: It's because of line length limit. I will rework for sure to be more specific here.

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
daron666: I think that this one for sure should be protected

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
daron666: Yeah. Fair point. I'm going to make them protected.

src/main/scala/scorex/core/network/PeerConnectionHandler.scala
daron666: I agree about direction, but not sure that handshake should be here.

daron666:
@ceilican @kushti review please.

kushti:
@daron666 this PR is too big to review, could you please split it into different PRs, like the approach with WartRemover you can find in the PRs list?

Issue #176, opened 2018-02-08
Add WartRemover checks


EzequielPostan:
Would it be a good idea to add checks using wartremover? 

https://github.com/wartremover/wartremover

I have seen some good checks like 
- Remove unsafe calls like `head`, `last`, `get`, etc.
- Remove `var`s and mutable state
- Detect `Product` type inferences
- Remove the use of `Any`

and many others.
We could define a full list of desired checks from here: http://www.wartremover.org/doc/warts.html

ceilican:
I think this is generally a good idea, although I don't agree with a few built-in warts or with the suggested ways to remove some warts.

greenhat:
I found wartremover super useful. @EzequielPostan nice list, `Recursion` wart deserves a mention too. It forces you to think about stack safety in every recursion.

EzequielPostan:
Thanks for the comment @greenhat , I will go back to work on this issue soon.

@ceilican 
We can select warts we would like to adopt and ignore the rest. I also don't like some of the suggested ways to remove some warts, but we can apply our own style.
The good thing is that breaking the wart rules won't let you compile, so it enforces good practices.

I will enable some warts in a new branch and check how many issues it detects.

EzequielPostan:
@ceilican
I enabled `TraversableOps`  in a local branch and the plugin found the following error cases:

```
sbt:scorex-core> test:compile
···
[error] .../src/main/scala/scorex/core/api/http/CompositeHttpService.scala:21:44: [wartremover:TraversableOps] reduce is disabled - use reduceOption or fold instead
[error]   val compositeRoute = routes.map(_.route).reduce(_ ~ _) ~ corsHandler(swaggerService.route) ~
[error]                                            ^
[error] .../src/main/scala/scorex/core/consensus/BlockChain.scala:39:36: [wartremover:TraversableOps] head is disabled - use headOption instead
[error]   def lastBlock: B = lastBlocks(1).head
[error]                                    ^
[error] .../src/main/scala/scorex/core/consensus/BlockChain.scala:50:29: [wartremover:TraversableOps] head is disabled - use headOption instead
[error]     val modId = openSurface.head._1
[error]                             ^
[error] .../src/main/scala/scorex/core/consensus/BlockChain.scala:51:37: [wartremover:TraversableOps] head is disabled - use headOption instead
[error]     val s = lookForward(openSurface.head._2, size)
[error]                                     ^
[...]

PR #188, opened 2018-02-16
I-176 add wartremover checks (TraversableOps enabled)

EzequielPostan:
This PR shows the errors pointed by the WartRemover plugin using the `TraversableOps` wart as commented in #176 

The results show that the code is currently unsafe in some parts. 
During the implementation I noticed other unsafe practices such as using `get` on Option and Try instances.

### DIFF
--- a/examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
+++ b/examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
@@ -318,7 +318,7 @@ class HybridHistory(val storage: HistoryStorage,
         log.warn(s"CompareNonsense: ${other.lastPowBlockIds.toList.map(Base58.encode)} at height $height}")
         HistoryComparisonResult.Nonsense
       case 1 =>
-        // TODO: What should happen if `dSuffix` is empty?
+        // `dSuffix.head` is safe as `dSuffix.length` is 1
         @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
         val dSuffixHead = dSuffix.head
         if (dSuffixHead sameElements bestPowId) {
--- a/examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
+++ b/examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
@@ -194,8 +194,12 @@ class HybridHistory(val storage: HistoryStorage,
 
     val rollbackPoint = newSuffix.headOption
 
-    val throwBlocks = oldSuffix.drop(1).map(id => modifierById(id).get)
-    val applyBlocks = newSuffix.drop(1).map(id => modifierById(id).get) ++ Seq(block)
+    // TODO: fixme, What should we do if `oldSuffix` is empty?
+    @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
+    val throwBlocks = oldSuffix.tail.map(id => modifierById(id).get)
+    // TODO: fixme, What should we do if `newSuffix` is empty?
+    @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
+    val applyBlocks = newSuffix.tail.map(id => modifierById(id).get) ++ Seq(block)
     require(applyBlocks.nonEmpty)
     require(throwBlocks.nonEmpty)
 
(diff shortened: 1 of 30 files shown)


examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
```diff
@@ -407,14 +417,25 @@ class HybridHistory(val storage: HistoryStorage,
...
     val winnerChain = chainBack(forkBlock, in, limit).get.map(_._2)
-    val i = loserChain.indexWhere(id => id sameElements winnerChain.head)
+    val i = loserChain.indexWhere { id =>
+      winnerChain.headOption match {
+        case None                  => false
+        case Some(winnerChainHead) => id sameElements winnerChainHead
+      }
+    }
     (winnerChain, loserChain.takeRight(loserChain.length - i))
-  }.ensuring(r => r._1.head sameElements r._2.head)
```
ceilican: Here, for example, both the original code and the refactored code will throw an exception when `r._1.head` or `r._2.head` is undefined. In the original code, this will occur because `head` is undefined. In the refactored code this will happen later because the condition will evaluate to false and `ensuring` will throw an exception. So, both codes achieve the same effect, but the original code is:

1. simpler and more concise
2. more efficient (throwing the exception earlier)
3. easier to understand for readers who don't know `Option`, `for` notation and `getOrElse`.

kushti:
@EzequielPostan please resolve the conflicts

examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
```diff
@@ -407,14 +424,26 @@ class HybridHistory(val storage: HistoryStorage,
     def in(m: HybridBlock): Boolean = loserChain.exists(s => s sameElements m.id)
 
     val winnerChain = chainBack(forkBlock, in, limit).get.map(_._2)
-    val i = loserChain.indexWhere(id => id sameElements winnerChain.head)
+    val i = loserChain.indexWhere { id =>
+      winnerChain.headOption match {
+        case None                  => false
+        case Some(winnerChainHead) => id sameElements winnerChainHead
+      }
+    }
```
EzequielPostan: I made this change here as I think it is the intended semantics. Should I take it back?

examples/src/main/scala/examples/spv/KLS16Proof.scala
```diff
@@ -71,8 +78,8 @@ object KLS16ProofSerializer extends Serializer[KLS16Proof] {
       Bytes.concat(Shorts.toByteArray(bytes.length.toShort), bytes)
     })
     Bytes.concat(Array(obj.m.toByte, obj.k.toByte, obj.i.toByte),
-      Shorts.toByteArray(obj.suffix.head.bytes.length.toShort),
-      obj.suffix.head.bytes,
+      Shorts.toByteArray(obj.suffix.headOption.map(_.bytes.length.toShort) getOrElse 0),
+      obj.suffix.headOption.map(_.bytes) getOrElse Array[Byte](),
```
EzequielPostan: Is this change correct?

examples/src/main/scala/examples/spv/KMZProof.scala
```diff
@@ -36,8 +38,8 @@ object KMZProofSerializer extends Serializer[KMZProof] {
     })
 
     Bytes.concat(Array(obj.m.toByte, obj.k.toByte),
-      Shorts.toByteArray(obj.suffix.head.bytes.length.toShort),
-      obj.suffix.head.bytes,
+      Shorts.toByteArray(obj.suffix.headOption.map(_.bytes.length.toShort) getOrElse 0),
+      obj.suffix.headOption.map(_.bytes) getOrElse Array[Byte](),
```
EzequielPostan: Same here, is this change correct?

examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
ceilican: I also think this is the intended semantics.

examples/src/main/scala/examples/spv/KLS16Proof.scala
ceilican: I think so. @kushti, could you double check?

ceilican: Actually, I'm not so sure... After all, having no head is not necessarily the same thing as having a head that is an empty array of bytes. And your code identifies these two possibly different things.

examples/src/main/scala/examples/spv/KMZProof.scala
ceilican: Same remark as above.

examples/src/main/scala/examples/spv/KLS16Proof.scala
EzequielPostan: Ok, I can take this change back and the one below too.
Besides this, I have no more comments. I will revert this change and send the final PR to review.
Thanks @ceilican :)

examples/src/main/scala/examples/spv/KMZProof.scala
EzequielPostan: same answer as above

kushti:
@EzequielPostan Thanks for the contributions!

Issue #178, opened 2018-02-09
Wrapper for Array[Byte]


catena2w:
Array[Byte] is bad as a key in maps/collections, so we anyway wrap them when use. It's better to wrap them once at object creation time. 
IODB requires ByteArrayWrapper as a key and value, may be we should wrap at least some objects (e.g. ModifierId) at object creation time

TharinduDG:
@catena2w 
Can this be closed since the requirement changed and there's a new issue for this? 
https://github.com/ScorexFoundation/Scorex/issues/208

PR #193, opened 2018-02-20
Wrapper for byte array

TharinduDG:
[ISSUE-178](https://github.com/ScorexFoundation/Scorex/issues/178)


catena2w:
Please:
1. Resolve conflicts
2. Define equals
3. Define hashCode
4. Define toString

TharinduDG:
@catena2w Done.
I had to use `!@@` in some places. Please check if its ok or not.

kushti:
@TharinduDG please resolve the conflicts!

TharinduDG:
@kushti Done. Please check.

examples/src/main/scala/examples/hybrid/validation/ParentBlockValidator.scala
```diff
@@ -14,16 +15,16 @@ class ParentBlockValidator(storage: HistoryStorage)
...
       case powBlock: PowBlock => if (!storage.isGenesis(powBlock)) {
         //check PoW parent id ???
-        require(storage.modifierById(powBlock.parentId).isDefined, s"Parent ${Base58.encode(powBlock.parentId)} missed")
+        require(storage.modifierById(ModifierId !@@ powBlock.parentId.arr).isDefined, s"Parent ${Base58.encode(powBlock.parentId.arr)} missed")
         //check referenced PoS block exists as well
         val posBlock = storage.modifierById(powBlock.prevPosId).get
 
         //check referenced PoS block points to parent PoW block
-        require(posBlock.parentId sameElements posBlock.parentId, "ref rule broken")
+        require(posBlock.parentId.arr sameElements posBlock.parentId.arr, "ref rule broken")
```
kushti: should be "posBlock.parentId == posBlock.parentId" , as a particular motivation to do boxing was to have "==" operator working (please check all other sameElements occurences, there are a lot of them!)

kushti:
@ceilican arrays (of primitive types) are about much less memory footprint, see http://www.lihaoyi.com/post/BenchmarkingScalaCollections.html#arrays-are-great . maybe WrappedArray is the solution

ceilican:
@kushti , `WrappedArray` is a `Seq`. So, your suggestion is a special case of my suggestion.

Note that we don't need to use `WrappedArray` directly. Whenever we have an `Array` and a `Seq` is expected, an implicit conversion from `Array` to `WrappedArray` will occur.  (Of course, though, we might want to use `WrappedArray` directly to minimize the number of times that such implicit conversions occur.)

My main point is: I think we should use `Seq` instead of `Array`, because `Array` is mutable, bug-prone, java-ish, ...  Which type of `Seq` (e.g. `WrappedArray`, `List`, `Vector`,...) doesn't matter so much to me.

So, what if we do the following:
1. When we receive raw data from another peer, we create a `WrappedArray` of bytes, instead of an `Array` of bytes. (as you suggested.)
2. Everywhere else in the code, where the data is consumed, we use the more general `Seq` (of which `WrappedArray` is just a particular subclass).

This would mean that, everywhere else in the code, we are not committing ourselves to any particular implementation of `Seq`. In the future, we could even experiment changing `WrappedArray` to `Vector` or even to our own optimized implementation of `Seq`, if we wish.

ceilican:
@kushti, what do you think?

kushti:
@ceilican while WrappedArray is implementing Seq, under the hood it is just a wrapper for an Array, see Wrapped.ofByte class definition. 

Around the project, we're using ByteStr, ByteString(from Akka), ByteBuffer. There's also ByteArrayWrapper in IODB (which is basically WrappedArray, but with faster comparison). I'm for any option (WrappedArray or maybe ByteString), but lets' be consistent around the code.

ceilican:
@kushti 

> @ceilican while WrappedArray is implementing Seq, under the hood it is just a wrapper for an Array, see Wrapped.ofByte class definition.

> Around the project, we're using ByteStr, ByteString(from Akka), ByteBuffer. There's also ByteArrayWrapper in IODB (which is basically WrappedArray, but with faster comparison). I'm for any option (WrappedArray or maybe ByteString), but lets' be consistent around the code.

Ok. I will write an issue about this. 

What should we do this PR then? I think it will become obsolete once we replace all `Array[Byte]` by, let's say,`ByteString`.  Should we close it?

kushti:
@ceilican yes, I think this it is better to close this PR

Issue #202, opened 2018-03-11
Refactor subscription mechanism


ceilican:
How important is the subscription pattern in Scorex?

I noticed that we frequently have the following:

* an actor `A` receives a message `M` and notifies every subscriber `S` (which subscribes to the message `M`) that an event `EventType.M` occurred.

Now in cases where `A` knows in advance who the subscriber `S` is, it would be more direct and easier to simply let `A` send a message to `S` when `A` receives `M`, instead of using the more indirect subscription mechanism.

I suppose the subscription pattern is needed because `A` doesn't always know in advance who the subscribers will be. But I don't see clearly where this is the case. @kushti, could you tell me why/where subscriptions are really needed? Could you point me to a place in Scorex's code where the more direct alternative would not work?

daron666:
Akka has embedded message bus, that could be used for such things, btw.

ceilican:
Thanks, @daron666 !

Here is a link to Akka's documentation about the event bus: https://doc.akka.io/docs/akka/2.5/event-bus.html

I think we should eliminate subscriptions as much as possible and use Akka's event bus in cases where subscriptions are really needed.

kushti:
@ceilican what concrete cases do you have in your mind? Sometimes message-passing through an intermediate actor done intentionally in order not to reveal unnecessary interfaces or pass unneeded dependencies.

ceilican:
@kushti , the issue here is not the use of intermediate actors. The issue here is the use of a complex "subscription pattern" just to achieve communication between two actors `A` and `S`, which could probably be achieved in a simpler way. For a concrete case, consider the following:

The *peer manager* (`A`) needs to tell the *node view synchronizer* (`S`) that a *handshake* (`M`) occurred. 

Right now, `S` first needs to send a message to `A` subscribing to messages of type `M`. Then later `A` sends message `M` to all subscribers, even though `S` is the **only** subscriber.

Wouldn't it be simpler to simply make `A` send `M` to `S`, without requiring `S` to subscribe first?

I see your point that, without the subscription mechanism, we would need to have an explicit dependency of `A` on `S`, but I think that in cases where there is a single subscriber, it can be an advantage to have an explicit dependency.  For example, if we look at the code of *peer manager* now, we have no idea who the subscribers are. With an explicit dependency, on the other hand, it would be clear that there is communication between the peer manager and the node view synchronizer.

[...]

Issue #197, opened 2018-03-02
Replace Enumerations by Case Objects


ceilican:
Scorex has 4 Enumeration objects: `CommunicationState`, `EventType`, `ModifiersSemanticValidity` and `HistoryComparisonResult`. I think it would be more elegant to implement them with case objects instead. Like this:

Instead of:

```
  object EventType extends Enumeration {
    val Handshaked: EventType.Value = Value(1)
    val Disconnected: EventType.Value = Value(2)
  }
```

We would do:

```
sealed trait EventType
case object Handshaked extends EventType
case object Disconnected extends EventType
```

@kushti: what do you think?

kushti:
@ceilican I agree, there're some pros of sealed traits

EzequielPostan:
@ceilican @kushti 

This is actually the purpose of Wart Enumeration (http://www.wartremover.org/doc/warts.html)
Would you like me to take this ticket and add this wart in the process?

ceilican:
@EzequielPostan , I have already done most of it in #203 .  There is only one enumeration remaining, but it would be useful to decide what to do with #202 before eliminating it.

ceilican:
@EzequielPostan , if you would like to eliminate the last remaining enumeration, please go ahead. Base your PR on top of #203 . Note that there will be some name clashes when you remove that enumeration. Try to rename the clashing events sensibly.

EzequielPostan:
@ceilican 
Ok, I will finish with the current work I have in a branch for #176 and then will check this issue.

ceilican:
@EzequielPostan , let's put this issue on hold, because it would be in conflict with #202 .  You haven't started working on this yet, right?

EzequielPostan:
No problem @ceilican, I haven't started yet.

PR #203, opened 2018-03-11
Replace Enumerations by Sealed Traits and Case Objects

ceilican:
Addresses issue #197 .

All enumerations have been replaced, except the one for NodeViewHolder's `EventType`. This last remaining enumeration is quite long, and it would be a good idea to discuss #202 before replacing it.

@kushti, only after working on this I have seen that you had tagged #197 with "help wanted". Sorry for that.


examples/src/main/scala/examples/hybrid/HybridApp.scala
```diff
@@ -58,11 +58,6 @@ class HybridApp(val settingsFilename: String) extends Application {
                                                      (networkControllerRef, nodeViewHolderRef, localInterface,
                                                       HybridSyncInfoMessageSpec, settings.network, timeProvider))
 
-  //touching lazy vals
```
kushti: why these lines are removed?

ceilican: Because the `val`s are not `lazy val`s anymore. So, there is no point in touching them. They are already touched when they are created.

kushti:
@ceilican please resolve conflicts!

ceilican:
@kushti, the conflicts have been resolved.

PR #210, opened 2018-03-21
Mempool transactions tests

anatoly-shamov:
Added some mempool tests. Also, I've added a mempool generator in MemoryPoolTest trait due to the usage of stateful mempool in hybrid example.


kushti:
@anatoly-shamov thanks for the contributions!

Issue #191, opened 2018-02-19
Node crash due to failed rollback in Hybrid


dkaidalov:
There is a problem happening all the time to me if I run 2 or more nodes. A node may crash with the following exception:
```  
java.util.NoSuchElementException: versionID not found, can not rollback
	at io.iohk.iodb.LSMStore.notFound$1(LSMStore.scala:868)
	at io.iohk.iodb.LSMStore.$anonfun$rollback$2(LSMStore.scala:875)
	at scala.Option.getOrElse(Option.scala:121)
	at io.iohk.iodb.LSMStore.rollback(LSMStore.scala:875)
	at examples.hybrid.state.HBoxStoredState.$anonfun$rollbackTo$1(HBoxStoredState.scala:88)
	at scala.util.Try$.apply(Try.scala:209)
	at examples.hybrid.state.HBoxStoredState.rollbackTo(HBoxStoredState.scala:84)
	at scorex.core.NodeViewHolder.$anonfun$updateState$1(NodeViewHolder.scala:230)
	at scala.util.Try$.apply(Try.scala:209)
	at scorex.core.NodeViewHolder.updateState(NodeViewHolder.scala:226)
	at scorex.core.NodeViewHolder.updateState$(NodeViewHolder.scala:220)
	at examples.hybrid.HybridNodeViewHolder.updateState(HybridNodeViewHolder.scala:26)
	at examples.hybrid.HybridNodeViewHolder.pmodModify(HybridNodeViewHolder.scala:123)
	at examples.hybrid.HybridNodeViewHolder.pmodModify(HybridNodeViewHolder.scala:26)
	at scorex.core.NodeViewHolder$$anonfun$processLocallyGeneratedModifiers$1.applyOrElse(NodeViewHolder.scala:373)
	......
```
So actually at some moment a node isn't able to do a rollback. A brief look at the issue brought me to this function [source code link](https://github.com/ScorexFoundation/Scorex/blob/master/examples/src/main/scala/examples/hybrid/history/HybridHistory.scala#L188-L204) which, as I can see, isn't fully implemented. It is probably the cause of the crashes (not fully sure though).

Are you also having this problem? Are you going to fix it anytime soon?

ceilican:
@dkaidalov , I have recently executed a few nodes (see #152), and I didn't experience this behaviour. Could you tell us in more detail what you did to see this error?

daron666:
[...]

PR #211, opened 2018-03-24
Fix for #191

kushti:
Forks resolution implementation for the Twinscoin prototype impl

### DIFF
--- a/examples/src/main/scala/examples/hybrid/HLocalInterface.scala
+++ b/examples/src/main/scala/examples/hybrid/HLocalInterface.scala
@@ -41,7 +41,7 @@ class HLocalInterface(override val viewHolderRef: ActorRef,
   override protected def onSemanticallySuccessfulModification(mod: HybridBlock): Unit = {
     if (!block) {
       mod match {
-        case wb: PowBlock =>
+        case _: PowBlock =>
           posForgerRef ! StartForging
           powMinerRef ! MineBlock
 
--- a/examples/src/main/scala/examples/hybrid/history/HistoryStorage.scala
+++ b/examples/src/main/scala/examples/hybrid/history/HistoryStorage.scala
@@ -1,18 +1,17 @@
 package examples.hybrid.history
 
 import com.google.common.primitives.Longs
-import examples.commons.SimpleBoxTransaction
 import examples.hybrid.blocks._
 import examples.hybrid.mining.{HybridMiningSettings, PosForger}
 import io.iohk.iodb.{ByteArrayWrapper, LSMStore}
 import scorex.core.ModifierId
-import scorex.core.block.Block
-import scorex.core.transaction.box.proposition.PublicKey25519Proposition
+import scorex.core.consensus.{Absent, ModifierSemanticValidity, Unknown}
 import scorex.core.utils.ScorexLogging
 import scorex.crypto.hash.Sha256
 
 import scala.util.Failure
 
+//TODO: why to use IODB if there's no rollback?
 class HistoryStorage(storage: LSMStore,
                      settings: HybridMiningSettings) extends ScorexLogging {
 
@@ -39,8 +38,7 @@ class HistoryStorage(storage: LSMStore,
     modifierById(bestPosId).get.asInstanceOf[PosBlock]
   }
 
-  def modifierById(blockId: ModifierId): Option[HybridBlock with
-    Block[PublicKey25519Proposition, SimpleBoxTransaction]] = {
+  def modifierById(blockId: ModifierId): Option[HybridBlock] = {
     storage.get(ByteArrayWrapper(blockId)).flatMap { bw =>
       val bytes = bw.data
       val mtypeId = bytes.head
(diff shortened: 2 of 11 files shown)


examples/src/main/scala/examples/hybrid/HLocalInterface.scala
```diff
@@ -41,7 +41,7 @@ class HLocalInterface(override val viewHolderRef: ActorRef,
   override protected def onSemanticallySuccessfulModification(mod: HybridBlock): Unit = {
     if (!block) {
       mod match {
-        case wb: PowBlock =>
+        case _: PowBlock =>
           posForgerRef ! StartForging
           powMinerRef ! MineBlock
```
ceilican: Maybe this comment is not relevant to this PR, but one thing that I find confusing here is that an expected symmetry seems to be broken. 

When we receive a PosBlock, we stop forging and start mining. I was expecting that we would stop mining and start forging when we receive a PowBlock. But, instead, what we see here is a command for the `powMinerRef` to mine a block instead of stop mining.

Could you explain this, perhaps as a comment in the code, if you consider appropriate?

examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
```diff
@@ -450,17 +465,24 @@ class HybridHistory(val storage: HistoryStorage,
     chainBack(storage.bestPosBlock, isGenesis).get.map(_._2).map(Base58.encode).mkString(",")
   }
 
-  //todo: real impl
   override def reportSemanticValidity(modifier: HybridBlock,
                                       valid: Boolean,
```
ceilican: Maybe it would be more elegant to change the type of `valid` to the superclass of `Valid` and `Invalid`...

examples/src/main/scala/examples/hybrid/history/HybridHistory.scala
```diff
@@ -450,17 +465,24 @@ class HybridHistory(val storage: HistoryStorage,
...
                                       lastApplied: ModifierId): (HybridHistory, ProgressInfo[HybridBlock]) = {
-    this -> ProgressInfo(None, Seq(), None, Seq())
-  }
+    val v = if (valid) Valid else Invalid
+    storage.updateValidity(modifier, v)
 
-  //todo: real impl
-  override def isSemanticallyValid(modifierId: ModifierId): ModifierSemanticValidity = {
```
ceilican: ... or change the type of `v` to `Boolean` in `storage.updateValidity`.

Just so that we don't have to convert from `Boolean` to `Valid` or `Invalid`.

examples/src/main/scala/examples/hybrid/history/HistoryStorage.scala
```diff
@@ -58,7 +56,28 @@ class HistoryStorage(storage: LSMStore,
...
+        .get(validityKey(b))
+        .map(_.data.head)
+        .map(ModifierSemanticValidity.restoreFromCode)
+        .getOrElse(Unknown)
+    }.getOrElse(Absent)
+  }
+
+  def updateValidity(b: HybridBlock, status: ModifierSemanticValidity) {
+    val version = ByteArrayWrapper(Sha256(b.id :+ status.code))
+    storage.update(version, Seq(), Seq(validityKey(b) -> ByteArrayWrapper(Array(status. code))))
```
ceilican: extra space after `.`.

src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
```diff
@@ -1,7 +1,30 @@
...
-case object Valid extends ModifierSemanticValidity
-case object Invalid extends ModifierSemanticValidity
\ No newline at end of file
+sealed trait ModifierSemanticValidity {
+  val code: Byte
+}
+
+object ModifierSemanticValidity {
+  def restoreFromCode(code: Byte): ModifierSemanticValidity = code match {
+    case b: Byte if b == Absent.code => Absent
```
ceilican: `if ... else if ... else` would be more succinct and cleaner than pattern matching here.

src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
```diff
@@ -1,7 +1,30 @@
...
-case object Unknown extends ModifierSemanticValidity
-case object Valid extends ModifierSemanticValidity
-case object Invalid extends ModifierSemanticValidity
\ No newline at end of file
+sealed trait ModifierSemanticValidity {
+  val code: Byte
+}
+
+object ModifierSemanticValidity {
+  def restoreFromCode(code: Byte): ModifierSemanticValidity = code match {
```
ceilican: maybe `restoreFromCode` ---> `fromCode` ?  `restore` does not add much information and makes the method name much longer.

src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
```diff
@@ -1,7 +1,30 @@
 package scorex.core.consensus
 
-sealed trait ModifierSemanticValidity
-case object Absent extends ModifierSemanticValidity
-case object Unknown extends ModifierSemanticValidity
-case object Valid extends ModifierSemanticValidity
-case object Invalid extends ModifierSemanticValidity
\ No newline at end of file
+sealed trait ModifierSemanticValidity {
+  val code: Byte
```
ceilican: This seems ok. But I wonder if it is the best approach. It introduces a dependency between our abstract data types and their low level representation in a block. I think we can accept this PR for now. Maybe add a "TODO" comment reminding ourselves to rethink this in the future. 

(I think that in general it might be a good idea to improve the decoupling of abstract data structures and their low-level representations. But this is something we could do in a future PR.)

src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
ceilican: And more efficient too.

PR #227, opened 2018-04-13
Re-fixed logic for outgoing connection to peers

### LINKED ISSUE #224: API peers/connect logic is broken
When we call `peers/connect` with a new peer address, node starts connecting given peer, but then it closes the connection as duplicate and doesn't add it to `connectedPeers`. The log message is following:

`Already connected peer ... trying to connect, going to drop the duplicate connection`

mike-aksarin:
Fixes #224. Re-fixed logic for outgoing connection to peers
(previous fix was incorrect and broke initial peer connections) 

I suggest to delete this pull request in favour of #229

### DIFF
--- a/src/main/scala/scorex/core/network/NetworkController.scala
+++ b/src/main/scala/scorex/core/network/NetworkController.scala
@@ -9,6 +9,10 @@ import akka.io.{IO, Tcp}
 import akka.pattern.ask
 import akka.util.Timeout
 import scorex.core.network.message.{Message, MessageHandler, MessageSpec}
+import NetworkController.ReceivableMessages._
+import NetworkControllerSharedMessages.ReceivableMessages.DataFromPeer
+import scorex.core.network.peer.PeerManager.ReceivableMessages._
+import PeerConnectionHandler.ReceivableMessages.CloseConnection
 import scorex.core.settings.NetworkSettings
 import scorex.core.utils.{NetworkTimeProvider, ScorexLogging}
 
@@ -30,11 +34,6 @@ class NetworkController(settings: NetworkSettings,
                         peerManagerRef: ActorRef,
                         timeProvider: NetworkTimeProvider
                        ) extends Actor with ScorexLogging {
-
-  import NetworkController.ReceivableMessages._
-  import NetworkControllerSharedMessages.ReceivableMessages.DataFromPeer
-  import scorex.core.network.peer.PeerManager.ReceivableMessages.{CheckPeers, FilterPeers, Disconnected}
-  import PeerConnectionHandler.ReceivableMessages.CloseConnection
 
   private implicit val system: ActorSystem = context.system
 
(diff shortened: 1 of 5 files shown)


mike-aksarin:
Let's delete this pull request in favour of https://github.com/ScorexFoundation/Scorex/pull/229

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -31,11 +35,6 @@ class NetworkController(settings: NetworkSettings,
                         timeProvider: NetworkTimeProvider
                        ) extends Actor with ScorexLogging {
 
-  import NetworkController.ReceivableMessages._
-  import NetworkControllerSharedMessages.ReceivableMessages.DataFromPeer
-  import scorex.core.network.peer.PeerManager.ReceivableMessages.{CheckPeers, FilterPeers, Disconnected}
-  import PeerConnectionHandler.ReceivableMessages.CloseConnection
```
EzequielPostan: @ceilican 
Should this imports be placed in the trait or moved to the top of the file like done here?

src/main/scala/scorex/core/network/UPnP.scala
```diff
@@ -13,8 +13,8 @@ class UPnP(settings: NetworkSettings) extends ScorexLogging {
 
   private var gateway: Option[GatewayDevice] = None
 
-  lazy val localAddress = gateway.map(_.getLocalAddress)
-  lazy val externalAddress = gateway.map(_.getExternalIPAddress).map(InetAddress.getByName)
+  lazy val localAddress: Option[InetAddress] = gateway.map(_.getLocalAddress)
+  lazy val externalAddress: Option[InetAddress] = gateway.map(_.getExternalIPAddress).map(InetAddress.getByName)
```
EzequielPostan: could this be simplified to a single `map`?

`gateway.map(InetAddress.getByName(_.getExternalIPAddress))`

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -180,7 +207,7 @@ object PeerManager {
     case object GetBlacklistedPeers
 
     // peerCycle messages
-    case class DoConnecting(remote: InetSocketAddress, direction: ConnectionType)
+    case class DoConnecting(remote: InetSocketAddress, direction: ConnectionType) // Only PeerHandler should send this
```
EzequielPostan: I wonder if this kind of restrictions could be enforced/checked in some way by the compiler

@ceilican @kushti 
any ideas on how we could do this?

EzequielPostan:
Oh, I first added some comments and then saw that this PR will be taken down.
My apologies for the notifications.

src/main/scala/scorex/core/network/NetworkController.scala
ceilican: I don't have a strong opinion about this. I just think that we should have a consistent style.

src/main/scala/scorex/core/network/peer/PeerManager.scala
ceilican: The only idea that comes to my mind is to use `protected[<PeerHandler's package...>]`, but this would require putting PeerHandler in its own package.

src/main/scala/scorex/core/network/NetworkController.scala
mike-aksarin: I agree about consistent style. I'd better move them to the top

src/main/scala/scorex/core/network/peer/PeerManager.scala
mike-aksarin: yes, you're right. But at least we can have a warn comment here.

PR #229, opened 2018-04-13
Fixing problems with outgoing connections dropping

kushti:


src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -102,14 +102,17 @@ class PeerManager(settings: ScorexSettings, timeProvider: NetworkTimeProvider) e
       if (peerDatabase.isBlacklisted(remote)) {
         log.info(s"Got incoming connection from blacklisted $remote")
       } else {
-        val peerHandlerRef = sender
-        val isIncoming = direction == Incoming
-        val isAlreadyConnecting = connectingPeers.contains(remote)
+        lazy val peerHandlerRef = sender
+        lazy val isIncoming = direction == Incoming
+        lazy val isAlreadyConnecting = connectingPeers.contains(remote)
```
mike-aksarin: looks like we don't need lazyness here always all three variables

PR #239, opened 2018-05-03
catch subscribed NodeViewChange events

catena2w:


ceilican:
@catena2w ,
I was planning to fix this issue in a different way. I would not subscribe NodeViewSynchronizer to vault and state modifications.

My understanding, from discussions with @kushti in Slack, was that vault and state modification events were supposed to be received and handled not by NodeViewSynchronizer but by another actor of the application.

Should I still go ahead and refactor this fix, after this PR is accepted?
Another thing I wanted to do is to move event messages to the same file of the actor that emits them. Right now, event messages are located in the same file of the actor that is supposed to receive them, but this is problematic in cases such as vault and state modification events, where we do not know the actor that should receive them.

catena2w:
@ceilican you can fix it in your way

PR #244, opened 2018-05-14
ModifierValidation facility

mike-aksarin:

PR #246, opened 2018-05-15
ApiResponse automatic json conversion and Null handling

mike-aksarin:
This includes #244 ModifierValidation  facility changes

### DIFF
--- a/src/main/scala/scorex/core/api/http/ApiResponse.scala
+++ b/src/main/scala/scorex/core/api/http/ApiResponse.scala
@@ -37,8 +37,8 @@ object ApiResponse {
   def apply[R: Encoder](result: R): Route = OK(result)
   def apply[R: Encoder](result: Future[R]): Route = OK(result)
   def apply(result: Either[Throwable, Json]): Route = result.fold(ApiError.apply, OK.apply)
-  def apply(keyValues: (String, Json)*): Route = apply(Map(keyValues: _*).asJson)
-  def apply(keyValue: (String, String)): Route = apply(Map(keyValue).asJson)
+  def apply(keyValues: (String, Json)*): Route = apply(Map(keyValues: _*))
+  def apply(keyValue: (String, String)): Route = apply(Map(keyValue))
 
   object OK extends ApiResponse(StatusCodes.OK)
 }
--- a/src/main/scala/scorex/core/api/http/ApiResponse.scala
+++ b/src/main/scala/scorex/core/api/http/ApiResponse.scala
@@ -2,8 +2,8 @@ package scorex.core.api.http
 
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, StatusCode, StatusCodes}
 import akka.http.scaladsl.server.{Directives, Route}
-import io.circe.{Encoder, Json}
 import io.circe.syntax._
+import io.circe.{Encoder, Json}
 
 import scala.concurrent.Future
 import scala.language.implicitConversions
@@ -25,9 +25,13 @@ class ApiResponse(statusCode: StatusCode) {
 
   def withJson[R](result: R)(implicit encoder: Encoder[R]): Route = complete(encoder(result))
 
-  def complete(result: Json): Route = result match {
-    case json if json.isNull => ApiError.NotExists
-    case _ => Directives.complete(statusCode.intValue() -> HttpEntity(ContentTypes.`application/json`, result.spaces2))
+  def complete(result: Json): Route = {
+    if (result.isNull) {
+      ApiError.NotExists
+    } else {
+      val httpEntity = HttpEntity(ContentTypes.`application/json`, result.spaces2)
+      Directives.complete(statusCode.intValue() -> httpEntity)
+    }
   }
 }
 
--- a/src/main/scala/scorex/core/api/http/ApiResponse.scala
+++ b/src/main/scala/scorex/core/api/http/ApiResponse.scala
@@ -2,7 +2,7 @@ package scorex.core.api.http
 
 import akka.http.scaladsl.model.{ContentTypes, HttpEntity, StatusCode, StatusCodes}
 import akka.http.scaladsl.server.{Directives, Route}
-import io.circe.Json
+import io.circe.{Encoder, Json}
 import io.circe.syntax._
 
 import scala.concurrent.Future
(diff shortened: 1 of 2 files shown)


src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
```diff
@@ -10,20 +10,21 @@ object ModifierSemanticValidity {
     else if (code == Unknown.code) Unknown
     else if (code == Invalid.code) Invalid
     else Absent
-}
 
-case object Absent extends ModifierSemanticValidity {
-  override val code = 0
-}
+  case object Absent extends ModifierSemanticValidity {
+    override val code = 0.toByte
```
EzequielPostan: It may be better to explicitly state the type of `code` and avoid the `toByte` method.
```
override val code: Byte = 0
```
same for similar definitions below.

src/main/scala/scorex/core/api/http/ApiResponse.scala
```diff
@@ -2,35 +2,43 @@ package scorex.core.api.http
...
   def withString(s: String): Route = complete(s.asJson)
-  def withJson(result: Json): Route = complete(result)
 
-  def complete(result: Json): Route = {
-    Directives.complete(statusCode.intValue() -> HttpEntity(ContentTypes.`application/json`, result.spaces2))
+  def withJson[R](result: R)(implicit encoder: Encoder[R]): Route = complete(encoder(result))
+
+  def complete(result: Json): Route = result match {
+    case json if json.isNull => ApiError.NotExists
+    case _ => Directives.complete(statusCode.intValue() -> HttpEntity(ContentTypes.`application/json`, result.spaces2))
```
EzequielPostan: minor comment, the match expression could be replaced by an if expression
```
if(result.isNull) { 
  ApiError.NotExists
} else {
  Directives.complete(statusCode.intValue() -> HttpEntity(ContentTypes.`application/json`, result.spaces2))
}
```

src/main/scala/scorex/core/validation/ValidationResult.scala
```diff
@@ -0,0 +1,68 @@
...
+      next match {
+        case Valid => this
+        case Invalid(e2) => Invalid(errors ++ e2)
+      }
+    }
+
+    @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
+    lazy val toTry: Try[Unit] = {
+      if (errors.size == 1) Failure(errors.head.toThrowable) else Failure(MultipleErrors(errors))
+    }
```
EzequielPostan: In this definition I wonder how to report what I guess is an inconsistent state where `errors` is empty.
Also, do we need to separate the case of a single error and multiple errors? Can't we rename and use `MultipleErrors` for both scenarios?

src/main/scala/scorex/core/api/http/ApiResponse.scala
```diff
@@ -2,35 +2,43 @@ package scorex.core.api.http
...
   def defaultRoute: Route = withString(defaultMessage)
   def defaultMessage: String = statusCode.reason
 
   def withString(s: String): Route = complete(s.asJson)
-  def withJson(result: Json): Route = complete(result)
 
-  def complete(result: Json): Route = {
-    Directives.complete(statusCode.intValue() -> HttpEntity(ContentTypes.`application/json`, result.spaces2))
+  def withJson[R](result: R)(implicit encoder: Encoder[R]): Route = complete(encoder(result))
+
```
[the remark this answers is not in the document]
mike-aksarin: we won't have `encoder` val then and will have to write implicitly[Encoder[R]].apply

src/main/scala/scorex/core/validation/ValidationResult.scala
mike-aksarin: we could use `NonEmptyList` from cats in `MultipleErrors` but I don't really think that we need this complication. If somebody accidentally creates MultipleErrors with empty list, well this will not have any fatal consequences.

I'm not wrapping single error to `MultipleErrors` class to make the stacktrace more clean. It's easier to see the cause of an exception immediately in log, not looking for nested exception.

As for making all error Multiple i don't thing it's a good idea, because in this case we could have something looking like tree of exceptions inside, not a list

src/main/scala/scorex/core/consensus/ModifierSemanticValidity.scala
mike-aksarin: This works, thank you! I didn't notice that we are missing type annotation here.

src/main/scala/scorex/core/api/http/ApiResponse.scala
EzequielPostan: I see, I would favor the if expression based on this discussion (mainly efficiency) https://stackoverflow.com/questions/9266822/pattern-matching-vs-if-else
I personally also prefer it on style, but my taste is not a good reason for changing this.

@kushti @ceilican 
is there any reason why we see/prefer pattern matching over if expressions as @mike-aksarin  noticed in the code?

src/main/scala/scorex/core/api/http/ApiResponse.scala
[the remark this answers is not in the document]
EzequielPostan: It's ok, this is less verbose as you remark.

src/main/scala/scorex/core/validation/ValidationResult.scala
EzequielPostan: I agree that using `NonEmptyList` would be too much for this. I initially focused on this snippet when I saw the `SuppressWarnings` annotation. I think that it will be better to keep the logic simple as you stated rather than focus too much on the empty list detail.
Thank you for the time to reply to my comments.

ceilican: It looks like we are reinventing the wheel here. Couldn't we use cats `Validated` (https://typelevel.org/cats/datatypes/validated.html) to accumulate errors?

mike-aksarin: I've started this piece from trying to apply Cats Validated for modifier validation in Ergo, but it didn't suit. It's supposed to be used mostly for the web from validation. It's really good for accumulating all the validated fields and after that constructing a composite object from all these fields. 

We have a pretty different case, because we need to perform multiple checks for the same object without any transformation. This looks too messy with cats. See the example of that kind of validation in Ergo HeaderValidator here: https://github.com/ergoplatform/ergo/blob/i269_post_transatction/src/main/scala/org/ergoplatform/nodeView/history/storage/modifierprocessors/HeadersProcessor.scala#L300.

Also cats `Validated` doesn't support fail-fast validation, but we do.

ceilican: Ok. Thanks for explanation, @mike-aksarin !

I would suggest adding your explanation as a comment in the code. Otherwise other people will wonder why cats's `Validated` wasn't used.

mike-aksarin: Done!

PR #247, opened 2018-05-18
NetworkTimeProvider fix

catena2w:
- Do not allow concurrent updates
- Execute in a separate thread
- close https://github.com/ergoplatform/ergo/issues/281

### DIFF
--- a/src/main/scala/scorex/core/app/Application.scala
+++ b/src/main/scala/scorex/core/app/Application.scala
@@ -14,7 +14,7 @@ import scorex.core.transaction.box.proposition.Proposition
 import scorex.core.transaction.Transaction
 import scorex.core.utils.{NetworkTimeProvider, ScorexLogging}
 
-import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.ExecutionContext
 
 trait Application extends ScorexLogging {
 
@@ -36,6 +36,7 @@ trait Application extends ScorexLogging {
   implicit def exceptionHandler: ExceptionHandler = ApiErrorHandler.exceptionHandler
 
   protected implicit lazy val actorSystem = ActorSystem(settings.network.agentName)
+  implicit val executionContext: ExecutionContext = actorSystem.dispatchers.lookup("scorex.executionContext")
 
   protected val additionalMessageSpecs: Seq[MessageSpec[_]]
 
@@ -63,7 +64,7 @@ trait Application extends ScorexLogging {
     */
   val swaggerConfig: String
 
-  val timeProvider = new NetworkTimeProvider(settings.ntp)
+  val timeProvider = new NetworkTimeProvider(settings.ntp)(executionContext)
 
 
   val peerManagerRef = PeerManagerRef(settings, timeProvider)
--- a/src/main/scala/scorex/core/app/Application.scala
+++ b/src/main/scala/scorex/core/app/Application.scala
@@ -64,7 +64,7 @@ trait Application extends ScorexLogging {
     */
   val swaggerConfig: String
 
-  val timeProvider = new NetworkTimeProvider(settings.ntp)(executionContext)
+  val timeProvider = new NetworkTimeProvider(settings.ntp)
 
 
   val peerManagerRef = PeerManagerRef(settings, timeProvider)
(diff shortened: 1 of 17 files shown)


src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -16,60 +17,47 @@ object NetworkTime {
...
-    } catch {
-      case t: Throwable =>
-        log.warn("Problems with NTP: ", t)
-        None
-    } finally {
-      client.close()
-    }
+  private implicit val ec:ExecutionContext = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor)
+  private val lastUpdate = new AtomicLong(0)
+  private var offset = 0L
```
mike-aksarin: should be volatile or Atomic as could be updated from different threads

src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -16,60 +17,47 @@ object NetworkTime {
...
+          offset = newOffset
+          log.info("New offset adjusted: " + offset)
+          lastUpdate.set(time)
+        case Failure(e) =>
+          log.warn("Problems with NTP: ", e)
+          lastUpdate.set(lu)
+      }
+    } else {
+      // No update required. Set lastUpdate back to it's initial value
+      lastUpdate.set(lu)
```
mike-aksarin: should be lastUpdate.compareAndSet(time, lu)

src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -16,60 +17,47 @@ object NetworkTime {
...
+    if (time > lu + ntpSettings.updateEvery.toMillis) {
+      // time to update offset
+      updateOffSet().onComplete {
+        case Success(newOffset) =>
+          offset = newOffset
+          log.info("New offset adjusted: " + offset)
+          lastUpdate.set(time)
+        case Failure(e) =>
+          log.warn("Problems with NTP: ", e)
+          lastUpdate.set(lu)
```
mike-aksarin: should be lastUpdate.compareAndSet(time, lu)

src/main/scala/scorex/core/utils/NetworkTime.scala
```diff
@@ -16,60 +17,47 @@ object NetworkTime {
...
-      info.computeDetails()
-      Option(info.getOffset)
-    } catch {
-      case t: Throwable =>
-        log.warn("Problems with NTP: ", t)
-        None
-    } finally {
-      client.close()
-    }
+  private implicit val ec:ExecutionContext = ExecutionContext.fromExecutor(Executors.newSingleThreadExecutor)
```
mike-aksarin: Maybe to use akka thread-pool would be a better idea, though I'm not sure

src/test/scala/scorex/core/network/peer/PeerManagerSpec.scala
```diff
@@ -26,7 +26,7 @@ class PeerManagerSpec extends FlatSpec with Matchers {
     implicit val defaultSender: ActorRef = p.testActor
 
     val settings = ScorexSettings.read(None)
-    val timeProvider = new NetworkTimeProvider(settings.ntp)
+    val timeProvider = new NetworkTimeProvider(settings.ntp)(system.dispatcher)
```
mike-aksarin: Ohnoes! this was a really bad idea to use Akka's thread pull for the long operations! Sorry for that. Actually we need to introduce the special ExecutionContext for the time consuming operations, like Akka authors advise: https://doc.akka.io/docs/akka/2.5/dispatchers.html#Blocking_Needs_Careful_Management

mike-aksarin: We can even embrace the context lookup for these time-consuming tasks

`implicit val executionContext: ExecutionContext = context.system.dispatchers.lookup("my-blocking-dispatcher")`

src/main/scala/scorex/core/app/Application.scala
```diff
@@ -63,7 +64,7 @@ trait Application extends ScorexLogging {
     */
   val swaggerConfig: String
 
-  val timeProvider = new NetworkTimeProvider(settings.ntp)
+  val timeProvider = new NetworkTimeProvider(settings.ntp)(executionContext)
```
mike-aksarin: this is the only place, where we pass executionContext explicitly, other usages are implicit.

PR #254, opened 2018-06-12
Use flexible encoder for logs

catena2w:


src/main/scala/scorex/core/utils/ScorexLogging.scala
```diff
@@ -1,7 +1,11 @@
...
 
 import com.typesafe.scalalogging.StrictLogging
+import scorex.crypto.encode.{Base16, BytesEncoder}
 
-trait ScorexLogging extends StrictLogging {
+// TODO just extend from scrypto logging after the next release
+trait ScorexLogging  extends StrictLogging {
   @inline protected def log = logger
+
+  implicit val encoder: BytesEncoder = Base16
```
ceilican: I think encoding is unrelated to logging. I would suggest creating a `ScorexEncoding` trait and adding `implicit val encoder: BytesEncoder = Base16` there.

Issue #217, opened 2018-03-28
Make modifiersCache of limited size


kushti:
At the moment "modifiersCache" in NodeViewHolder can grow indefinitely. Configurable cache size & strategy to clean the cache are needed to be implemented.

ceilican:
Should we consider using a caching library (e.g. scaffeine, scalacache)?

PR #262, opened 2018-06-18
Reworked Modifiers Cache

kushti:
Fix for #217

### DIFF
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -133,8 +133,8 @@ scorex {
     # Network controller timeout
     controllerTimeout = 5s
 
-    # How many persistent modifiers to store in a cache.
-    # A modifier is waiting in the cache for being applied.
+    # How many persistent modifiers to store in the cache.
+    # The cache stores modifiers that are waiting to be applied.
     maxModifiersCacheSize = 256
   }
 
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -132,6 +132,10 @@ scorex {
 
     # Network controller timeout
     controllerTimeout = 5s
+
+    # How many persistent modifiers to store in a cache.
+    # A modifier is waiting in the cache for being applied.
+    maxModifiersCacheSize = 256
   }
 
   ntp {
--- a/src/main/resources/reference.conf
+++ b/src/main/resources/reference.conf
@@ -135,7 +135,7 @@ scorex {
 
     # How many persistent modifiers to store in the cache.
     # The cache stores modifiers that are waiting to be applied.
-    maxModifiersCacheSize = 256
+    maxModifiersCacheSize = 512
   }
 
   ntp {
--- a/src/main/scala/scorex/core/ModifiersCache.scala
+++ b/src/main/scala/scorex/core/ModifiersCache.scala
@@ -1,7 +1,5 @@
 package scorex.core
 
-
-import com.google.common.cache.CacheBuilder
 import scorex.core.consensus.HistoryReader
 import scorex.core.utils.ScorexLogging
 import scorex.core.validation.RecoverableModifierError
@@ -19,7 +17,7 @@ trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD
   type K = mutable.WrappedArray[Byte]
   type V = PMOD
 
-  protected val cache = mutable.Map[K, V]()
+  protected val cache: mutable.Map[K, V] = mutable.Map[K, V]()
 
   def size: Int = cache.size
 
@@ -35,7 +33,7 @@ trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD
     * to have this structure is to avoid repeatedly downloading modifiers
     * which are unquestionably invalid.
     */
-  protected val rememberedKeys = mutable.HashSet[K]()
+  protected val rememberedKeys: mutable.HashSet[K] = mutable.HashSet[K]()
 
   /**
     * Defines a best (and application-specific) candidate to be applied.
--- a/src/main/scala/scorex/core/ModifiersCache.scala
+++ b/src/main/scala/scorex/core/ModifiersCache.scala
@@ -4,6 +4,7 @@ import scorex.core.consensus.HistoryReader
 import scorex.core.utils.ScorexLogging
 import scorex.core.validation.RecoverableModifierError
 
+import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.util.{Failure, Success}
 
@@ -83,25 +84,32 @@ trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD
 
 trait LRUCache[PMOD <: PersistentNodeViewModifier, HR <: HistoryReader[PMOD, _]] extends ModifiersCache[PMOD, HR] {
 
-  type Counter = BigInt
+  private val evictionQueue = mutable.Queue[K]()
 
-  private var cnt = 0: Counter
+  // The eviction queue can contain elements already removed, as we're not removing a key from it when
+  // the key is got removed from the cache. When size of eviction queue exceeds maximum size of the cache by
+  // the value below(so the queue contains at least "cleaningThreshold" keys aleady removed from the cache),
+  // complete scan and cleaning of removed keys happen.
+  private val cleaningThreshold = 50
 
-  private val keyIndex = mutable.Map[K, Counter]()
+  @tailrec
+  private def evictionCandidate(): K = {
+    val k = evictionQueue.dequeue()
+    if(cache.contains(k)) k else evictionCandidate()
+  }
 
   override protected def onPut(key: K): Unit = {
-    keyIndex.put(key, cnt)
-    cnt = cnt + 1
+    evictionQueue.enqueue(key)
+    if(evictionQueue.size > maxSize + cleaningThreshold){
+      evictionQueue.dequeueAll(k => !cache.contains(k))
+    }
   }
 
   override protected def onRemove(key: K, rememberKey: Boolean): Unit = {
-    keyIndex.remove(key)
   }
 
   def keyToRemove(): K = {
-    val (key, _) = keyIndex.minBy(_._2)  //todo: can we do faster search?
-    keyIndex.remove(key)
-    key
+    evictionCandidate()
   }
 }
 
--- a/src/main/scala/scorex/core/ModifiersCache.scala
+++ b/src/main/scala/scorex/core/ModifiersCache.scala
@@ -89,7 +89,7 @@ trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD
 
 trait LRUCache[PMOD <: PersistentNodeViewModifier, HR <: HistoryReader[PMOD, _]] extends ModifiersCache[PMOD, HR] {
 
-  type Counter = Long
+  type Counter = BigInt
 
   private var cnt = 0: Counter
 
--- a/src/main/scala/scorex/core/ModifiersCache.scala
+++ b/src/main/scala/scorex/core/ModifiersCache.scala
@@ -1,0 +1,211 @@
+package scorex.core
+
+
+import scorex.core.consensus.History.ModifierIds
+import scorex.core.consensus.{History, HistoryReader, ModifierSemanticValidity, SyncInfo}
+import scorex.core.serialization.Serializer
+import scorex.core.utils.ScorexLogging
+import scorex.core.validation.RecoverableModifierError
+import scorex.crypto.hash.Blake2b256
+
+import scala.collection.mutable
+import scala.util.{Failure, Success, Try}
+
+
+trait ModifiersCache[PMOD <: PersistentNodeViewModifier] {
+  require(maxSize >= 1)
+
+  type H <: HistoryReader[PMOD, _]
+
+  type K = mutable.WrappedArray[Byte]
+  type V = PMOD
+
+  protected var currentCacheSize: Int = 0
+
+  protected val cache = mutable.Map[K, V]()
+
+  protected val rememberedKeys = mutable.HashSet[K]()
+
+  def findCandidateKey(history: H): Option[K]
+
+  protected def onPut(key: K): Unit = {}
+  protected def onRemove(key: K, rememberKey: Boolean): Unit = {}
+  protected def keyToRemove(): K
+
+
+  def contains(key: K): Boolean = cache.contains(key) || rememberedKeys.contains(key)
+
+  def put(key: K, value: V): Unit = synchronized {
+    onPut(key)
+    cache.put(key, value)
+    currentCacheSize = currentCacheSize + 1
+
+    if (currentCacheSize > maxSize) remove(keyToRemove())
+  }
+
+  // def get(key: K): Option[V] = cache.get(key)
+
+  def remove(key: K, rememberKey: Boolean = false): Option[V] = synchronized {
+    onRemove(key, rememberKey)
+    currentCacheSize = currentCacheSize - 1
+    if (rememberKey) rememberedKeys += key
+    cache.remove(key)
+  }
+
+  def popCandidate(history: H): Option[V] = synchronized {
+    findCandidateKey(history).flatMap(k => remove(k))
+  }
+
+  def size: Int = currentCacheSize
+
+  def maxSize: Int
+}
+
+trait LRUCache[PMOD <: PersistentNodeViewModifier] extends ModifiersCache[PMOD] {
+
+  type Counter = Long
+
+  private var cnt = 0: Counter
+
+  private val keyIndex = mutable.Map[K, Counter]()
+
+  override protected def onPut(key: K): Unit = {
+    keyIndex.put(key, cnt)
+    cnt = cnt + 1
+  }
+
+  override protected def onRemove(key: K, rememberKey: Boolean): Unit = {
+    keyIndex.remove(key)
+  }
+
+  @SuppressWarnings(Array("org.wartremover.warts.TraversableOps"))
+  def keyToRemove(): K = {
+    val (key, _) = keyIndex.minBy(_._2)  //todo: can we do faster search?
+    keyIndex.remove(key)
+    key
+  }
+}
+
+class DefaultModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD, _]]
+  (override val maxSize: Int) extends ModifiersCache[PMOD] with LRUCache[PMOD] with ScorexLogging {
+
+  @SuppressWarnings(Array("org.wartremover.warts.IsInstanceOf"))
+  override def findCandidateKey(history: H): Option[K] = {
+
+    cache.find { case (k, v) =>
+      history.applicableTry(v) match {
+        case Failure(e) if e.isInstanceOf[RecoverableModifierError] =>
+          // do nothing - modifier may be applied in future
+          false
+        case Failure(e) =>
+          // non-recoverable error - remove modifier from cache
+          // TODO blaklist peer who sent it
+          log.warn(s"Modifier ${v.encodedId} is permanently invalid and will be removed from cache", e)
+          remove(k, rememberKey = true)
+          false
+        case Success(_) =>
+          true
+      }
+    }.map(_._1)
+  }
+}
+
+object DefaultModifiersCache extends App {
+
+  class FakeModifier extends PersistentNodeViewModifier {
+    override def parentId: ModifierId = ???
+
+    override val modifierTypeId: ModifierTypeId = ModifierTypeId @@ (0: Byte)
+
+    override def id: ModifierId = ???
+
+    override type M = this.type
+
+    override def serializer: Serializer[FakeModifier.this.type] = ???
+  }
+
+  class FakeSyncInfo extends SyncInfo {
+    override def startingPoints: ModifierIds = ???
+
+    override type M = this.type
+
+    override def serializer: Serializer[FakeSyncInfo.this.type] = ???
+  }
+
+  class FakeHr extends HistoryReader[FakeModifier, FakeSyncInfo] {
+    /**
+      * Is there's no history, even genesis block
+      */
+    override def isEmpty: Boolean = ???
+
+    /**
+      * Whether a modifier could be applied to the history
+      *
+      * @param modifier - modifier to apply
+      * @return `Success` if modifier can be applied, `Failure(ModifierError)` if can not
+      */
+    override def applicableTry(modifier: FakeModifier): Try[Unit] = ???
+
+    /**
+      * Return modifier of type PM with id == modifierId
+      *
+      * @param modifierId - modifier id to get from history
+      * @return
+      */
+    override def modifierById(modifierId: ModifierId): Option[FakeModifier] = ???
+
+    /**
+      * Return semantic validity status of modifier with id == modifierId
+      *
+      * @param modifierId - modifier id to check
+      * @return
+      */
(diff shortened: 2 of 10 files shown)


src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  /**
+    * A cache element replacement strategy method, which defines a key to remove from cache when it is overfull
+    */
+  protected def keyToRemove(): K
+
+
+  def contains(key: K): Boolean = cache.contains(key) || rememberedKeys.contains(key)
+
+  def put(key: K, value: V): Unit = synchronized {
```
ceilican: Do we need to synchronize the whole method? Or just the incrementation of the cachesize?

If it would suffice to synchronize only the incrementation, then it is recommended to use AtomicLong instead (https://stackoverflow.com/questions/34877487/how-to-use-synchronized-in-scala).

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  /**
+    * Remove an element from the cache.
+    * @param key - modifier's key
+    * @param rememberKey - whether to remember the key as belonging to cache. E.g. invalid modifiers are
+    *                    to be remembered (for not to be requested from the network again).
+    * @return
+    */
+  def remove(key: K, rememberKey: Boolean = false): Option[V] = synchronized {
```
ceilican: The same here.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+class DefaultModifiersCache[PMOD <: PersistentNodeViewModifier, HR <: HistoryReader[PMOD, _]]
+  (override val maxSize: Int) extends ModifiersCache[PMOD] with LRUCache[PMOD] with ScorexLogging {
+
+  override type H = HR
+
+  /**
+    * Default implementation is just about to scan. Not efficient at all and should be probably rewritten in a
+    * concrete application.
+    *
+    * @param history - an interface to history which could be needed to define a candiate
```
ceilican: "candiate" --> "candidate"

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+    * @return - candidate if it is found
+    */
+  @SuppressWarnings(Array("org.wartremover.warts.IsInstanceOf"))
+  override def findCandidateKey(history: H): Option[K] = {
+
+    cache.find { case (k, v) =>
+      history.applicableTry(v) match {
+        case Failure(e) if e.isInstanceOf[RecoverableModifierError] =>
```
ceilican: You can do `Failure(e:RecoverableModifierError)`.

Then you can remove the `SuppressWarnings`  a few lines above.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+
+  protected var currentCacheSize: Int = 0
+
+  protected val cache = mutable.Map[K, V]()
+
+  protected val rememberedKeys = mutable.HashSet[K]()
+
+  /**
+    * Defines a best (and application-specific) candidate to be applied.
+    * @param history - an interface to history which could be needed to define a candiate
```
ceilican: "candiate"

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+    * @return - candidate if it is found
+    */
+  def findCandidateKey(history: H): Option[K]
```
ceilican: Maybe the Cache is not the best place to locate this method. Alternatively, this method could be  implemented in the application that uses the cache instead. This would make the cache more agnostic and "ignorant" about application-specific logic, and would allow the cache to just cache. Maybe we could even use a cache library then (e.g. https://github.com/blemale/scaffeine or https://cb372.github.io/scalacache/).

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+    */
+  def remove(key: K, rememberKey: Boolean = false): Option[V] = synchronized {
+    onRemove(key, rememberKey)
+    currentCacheSize = currentCacheSize - 1
+    if (rememberKey) rememberedKeys += key
+    cache.remove(key)
+  }
+
+  def popCandidate(history: H): Option[V] = synchronized {
+    findCandidateKey(history).flatMap(k => remove(k))
```
ceilican: `k => remove(k)` could be simplified to `remove`

ceilican: `remove` has side-effects. It looks a bit strange to use a function with side-effects inside `flatMap` (or `filter`, `map`, ...).

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: Why do we return `Option[V]` here? It seems that the only reason is to use `remove` inside `flatMap` below. I think this saves only a few lines of code, but makes the intent of the code less clear.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+    currentCacheSize = currentCacheSize + 1
+
+    if (currentCacheSize > maxSize) remove(keyToRemove())
+  }
+
+  /**
+    * Remove an element from the cache.
+    * @param key - modifier's key
+    * @param rememberKey - whether to remember the key as belonging to cache. E.g. invalid modifiers are
+    *                    to be remembered (for not to be requested from the network again).
```
ceilican: I suggest moving "invalid modifiers are to be remembered (for not to be requested from the network again)" to line 33.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,136 @@
...
+    keyIndex.put(key, cnt)
+    cnt = cnt + 1
+  }
+
+  override protected def onRemove(key: K, rememberKey: Boolean): Unit = {
+    keyIndex.remove(key)
+  }
+
+  def keyToRemove(): K = {
+    val (key, _) = keyIndex.minBy(_._2)  //todo: can we do faster search?
```
ceilican: > can we do faster search?

You could keep the minimum key stored in a `var`, and do the search and update this `var` only when `onPut` is called on the minimum key.

src/test/scala/scorex/core/DefaultModifiersCacheSpecification.scala
```diff
@@ -0,0 +1,120 @@
...
+    class FakeSyncInfo extends SyncInfo {
+      override def startingPoints: ModifierIds = ???
+
+      override type M = this.type
+
+      override def serializer: Serializer[FakeSyncInfo.this.type] = ???
+    }
+
+    class FakeHr extends HistoryReader[FakeModifier, FakeSyncInfo] {
+      /**
```
ceilican: I think you can remove comments from non-implemented methods in "fake" classes.  They are not going to be used by the tests anyway.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  def maxSize: Int
+
+  type K = mutable.WrappedArray[Byte]
+  type V = PMOD
+
+  protected var currentCacheSize: Int = 0
+
+  protected val cache = mutable.Map[K, V]()
+
+  protected val rememberedKeys = mutable.HashSet[K]()
```
catena2w: please add comments, what kind of kees does it keep, what is the maximum size, etc

src/main/scala/scorex/core/ModifiersCache.scala
catena2w: Agree with @ceilican.
In addition, looks like it won't be possible to implement `// TODO blaklist peer who sent it` inside a Cache (and it's not a correct place for it)

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  protected def onPut(key: K): Unit = {}
+  protected def onRemove(key: K, rememberKey: Boolean): Unit = {}
+
+  /**
+    * A cache element replacement strategy method, which defines a key to remove from cache when it is overfull
+    */
+  protected def keyToRemove(): K
+
+
+  def contains(key: K): Boolean = cache.contains(key) || rememberedKeys.contains(key)
```
catena2w: Is `|| rememberedKeys.contains` correct here? We actually do not have this modifier in a cache

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  def put(key: K, value: V): Unit = synchronized {
+    onPut(key)
+    cache.put(key, value)
+    currentCacheSize = currentCacheSize + 1
```
catena2w: What if we already have this modifier in the cache? In such a case we should not increment `currentCacheSize` and call `onPut`

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+    * Remove an element from the cache.
+    * @param key - modifier's key
+    * @param rememberKey - whether to remember the key as belonging to cache. E.g. invalid modifiers are
+    *                    to be remembered (for not to be requested from the network again).
+    * @return
+    */
+  def remove(key: K, rememberKey: Boolean = false): Option[V] = synchronized {
+    onRemove(key, rememberKey)
```
catena2w: Again, if we don't have this modifier in a cache, we should not call `onRemove` and decrease cacheSize.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  /**
+    * Remove an element from the cache.
+    * @param key - modifier's key
+    * @param rememberKey - whether to remember the key as belonging to cache. E.g. invalid modifiers are
+    *                    to be remembered (for not to be requested from the network again).
+    * @return
+    */
+  def remove(key: K, rememberKey: Boolean = false): Option[V] = synchronized {
```
catena2w: `remove` method may be protected

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+    if (rememberKey) rememberedKeys += key
+    cache.remove(key)
+  }
+
+  def popCandidate(history: H): Option[V] = synchronized {
+    findCandidateKey(history).flatMap(k => remove(k))
+  }
+}
+
+trait LRUCache[PMOD <: PersistentNodeViewModifier, HR <: HistoryReader[PMOD, _]] extends ModifiersCache[PMOD, HR] {
```
catena2w: Why can't we use LRUCache from some library? This should solve some problems, including `//todo: can we do faster search?`

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,132 @@
...
+  override def findCandidateKey(history: HR): Option[K] = {
+
+    cache.find { case (k, v) =>
+      history.applicableTry(v) match {
+        case Failure(e) if e.isInstanceOf[RecoverableModifierError] =>
+          // do nothing - modifier may be applied in future
+          false
+        case Failure(e) =>
+          // non-recoverable error - remove modifier from cache
+          // TODO blaklist peer who sent it
```
catena2w: This todo implementation become totally impossible from here. May be we should return more info from this method, including absolutelly incorrect modifiers.

src/test/scala/scorex/core/DefaultModifiersCacheSpecification.scala
catena2w: +1. And may be it's better to extract this helper classes to directory with mocks

src/test/scala/scorex/core/DefaultModifiersCacheSpecification.scala
```diff
@@ -0,0 +1,120 @@
...
+import scorex.crypto.hash.Blake2b256
+
+import scala.util.Try
+
+class DefaultModifiersCacheSpecification extends PropSpec
+  with PropertyChecks
+  with GeneratorDrivenPropertyChecks
+  with Matchers {
+
+  property("ignore adding self as a peer") {
```
catena2w: test name looks like copy-paste of different a test. Please provide a correct name, that describes tested properties

src/test/scala/scorex/core/DefaultModifiersCacheSpecification.scala
```diff
@@ -0,0 +1,120 @@
...
+import org.scalatest.{Matchers, PropSpec}
+import org.scalatest.prop.{GeneratorDrivenPropertyChecks, PropertyChecks}
+import scorex.core.consensus.{History, HistoryReader, ModifierSemanticValidity, SyncInfo}
+import scorex.core.consensus.History.ModifierIds
+import scorex.core.serialization.Serializer
+import scorex.crypto.hash.Blake2b256
+
+import scala.util.Try
+
+class DefaultModifiersCacheSpecification extends PropSpec
```
catena2w: Would like to see more tests here (and may be in a ScorexTestkit):
- Application of the same modifier twice
- Remove of modifier, that is not in a cache
- Addition of incorrect modifier id to `rememberedKeys`
- Application of modifier, with a key in `rememberedKeys`

src/main/scala/scorex/core/ModifiersCache.scala
kushti: That's the point, to simulate that remembered keys are in cache.

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @catena2w what library are you talking about?

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican mutable.Map is  not thread-safe AFAIK, and for the abstract onPut() it is better not to assume thread-safety as well.

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican we're just replicating Map behaviour. What are other options?

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican @catena2w then the abstract method will be located in NodeViewHolder, which is already overload with diverse functionalities to concretize in an application. On banning the peer, could we use an event bus to send a signal to the network layer? We have work here been abandoned on the way unfortunately.

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican since there's the second parameter in remove() (with a default value), simplification is not possible. On side-effects, what's your point exactly?

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican Made additional comments there

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican the problem is that the minimum key could be removed, and then we need to do linear search again.

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: You're right. Mutable map has `remove` (https://www.scala-lang.org/api/current/scala/collection/mutable/Map.html#remove(key:K):Option[V]), which indeed returns an `Option`...  I didn't know this method, because I usually use either `-` or `-=`, which do not return an `Option`...

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: Ok!

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: Maybe this logic is also something that should also be implemented outside, as everything else related to blacklisting...

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: Possibly https://github.com/blemale/scaffeine or https://cb372.github.io/scalacache/ .

kushti:
@ceilican @catena2w Thanks for the feedback. I've addressed the objections expressed in your comments. Please make another round of the review process!

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  type Counter = Long
+
+  private var cnt = 0: Counter
+
+  private val keyIndex = mutable.Map[K, Counter]()
+
+  override protected def onPut(key: K): Unit = {
+    keyIndex.put(key, cnt)
+    cnt = cnt + 1
```
ceilican: This will eventually overflow...

ceilican: And the overflow would suddenly cause an unexpected behaviour.

ceilican: You could solve this problem (and also the "faster search" problem) by putting the keys in a `Queue` with a finite maximum size and popping the head of the queue, instead of putting them in a map with counters and popping the key with the minimum counter.

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -50,11 +50,13 @@ trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD
   def contains(key: K): Boolean = cache.contains(key) || rememberedKeys.contains(key)
 
   def put(key: K, value: V): Unit = synchronized {
-    onPut(key)
-    cache.put(key, value)
-    currentCacheSize = currentCacheSize + 1
+    if(!contains(key)) {
+      onPut(key)
```
ceilican: Can there ever be a case when you might want `onPut` to be called even when the cache already contains the key?

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,143 @@
...
+
+  /**
+    * How many elements are to be stored in the cache
+    */
+  def maxSize: Int
+
+  type K = mutable.WrappedArray[Byte]
+  type V = PMOD
+
+  protected var currentCacheSize: Int = 0
```
ceilican: Why do you need this var?

Couldn't you simply use `cache.size` whenever you need to know the current cache size?

src/main/scala/scorex/core/ModifiersCache.scala
kushti: @ceilican the problem here is that we need to remove an arbitrary element from the queue (and this operation is to be used more often when getting the first element), and this operation is about linear time for Queue I guess. So we need for Queue alternative with sub-linear by-key removal and sublinear min-value search. Doable if we have two indexes (Map[K, Count] and Map[Count,K]), but then we are doubling memory consumption. Any smarter options?

src/main/scala/scorex/core/ModifiersCache.scala
kushti: done

kushti:
@ceilican @catena2w I've removed currentCacheSize variable and made counter in LRUCache of BigInt type. Please make another round of the review process!

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+  * A cache which is storing persistent modifiers not applied to history yet.
+  * @tparam PMOD - type of a persistent node view modifier (or a family of modifiers).
+  */
+trait ModifiersCache[PMOD <: PersistentNodeViewModifier, H <: HistoryReader[PMOD, _]] {
+  require(maxSize >= 1)
+
+
+  protected val cache = mutable.Map[K, V]()
```
catena2w: I remember we had a convension, that class should define type parameters first, then variables and functions at the end. In this trait order looks random

src/main/scala/scorex/core/ModifiersCache.scala
```diff
@@ -0,0 +1,139 @@
...
+  type V = PMOD
+
+  /**
+    * Keys to simulate objects residing a cache. So if key is stored here,
+    * the membership check (contains()) shows that the key is in the cache,
+    * but the value corresponding to the key is not stored. The motivation
+    * to have this structure is to avoid repeatedly downloading modifiers
+    * which are unquestionably invalid.
+    */
+  protected val rememberedKeys = mutable.HashSet[K]()
```
catena2w: Should not it have limited size?

kushti: @catena2w hard to say, in general case. Descendants are free to rewrite the default behavior, which is about to store (usually, 32) bytes for any invalid modifier, before a restart.

src/main/scala/scorex/core/ModifiersCache.scala
kushti: fixed

src/main/scala/scorex/core/ModifiersCache.scala
catena2w: scalacache + Google Guava might be the best choice, as soon as we already have it in dependencies.
Guava have 2 options: `maximumSize(1000)` and `expireAfterWrite(10, TimeUnit.MINUTES)` that might both be useful

src/main/resources/reference.conf
```diff
@@ -132,6 +132,10 @@ scorex {
 
     # Network controller timeout
     controllerTimeout = 5s
+
+    # How many persistent modifiers to store in the cache.
+    # The cache stores modifiers that are waiting to be applied.
+    maxModifiersCacheSize = 256
```
catena2w: Why this number? I think it should not be less then 500, because it will lead to synchronization problems when INV contains more ids, than maxModifiersCacheSize

kushti: @catena2w increased to 512

src/main/scala/scorex/core/ModifiersCache.scala
ceilican: > however, you must also ensure that queue is not becoming too big if caches overflows do happen very rare

Yes, that's a good point. Not sure how I would solve that.

kushti:
@ceilican @catena2w I've implemented a new LRU cache implementation, which is using Queue() along with a lazy cleaning procedure, as was proposed by @ceilican .

On using an external caching framework, I think that a cache using a framework could be implemented in addition to LRUCache / DefaultModifiersCache, not as a replacement of it. If you think that it is needed, please open a ticket. I think that an ability to provide affirmative membership test for some keys not actually stored is needed, thus we need to use some additional structure in addition, to, say, the cache solution of Guava.

kushti:
@ceilican I dont' see a constant with the number "512". About "50", I'm going to make an issue.

src/main/resources/reference.conf
```diff
@@ -130,6 +130,10 @@ scorex {
 
     # Network controller timeout
     controllerTimeout = 5s
+
+    # How many persistent modifiers to store in the cache.
+    # The cache stores modifiers that are waiting to be applied.
+    maxModifiersCacheSize = 512
```
ceilican: Ah, "512" is already in a configuration file... Sorry.

PR #263, opened 2018-06-18
Do not log modifier validation by default

catena2w:


src/main/scala/scorex/core/validation/ModifierValidator.scala
```diff
@@ -86,10 +86,7 @@ object ModifierValidator extends ScorexLogging  {
   def fatal(errorMessage: String): Invalid = invalid(MalformedModifierError(errorMessage))
 
   /** unsuccessful validation with a given error*/
-  def invalid(error: ModifierError): Invalid = {
-    log.info(error.info)
-    Invalid(Seq(error))
-  }
+  def invalid(error: ModifierError): Invalid = Invalid(Seq(error))
```
kushti: What's good in this change?

catena2w: When there are a lot of modifiers in cache, we check them all and see plenty of messages in logs

PR #273, opened 2018-06-30
Implementation of peer features. 

kushti:
This PR is proposing to add abstract peer "features" to be advertised during handshaking.
A list of peer features (or capabilities) contains static characteristics of a running node. As an example, "services" bitfield announced in during handshaking in the Bitcoin networking protocol ("https://en.bitcoin.it/wiki/Protocol_documentation#version") shows services supported by a peer (e.g. whether it can serve peers around with full blocks or UTXO lookups etc). As another example, a feature can list supported protocols for encryption.

A feature is rather abstract interface, from Scorex core we require from a feature to be (de)serializable, and we require an Application trait implementation to define a concrete instantiation of peer features (to be then serialized and sent other wire to another peer).

In addition, this PR is introducing explicit handshaking message size limit.


src/main/scala/scorex/ObjectGenerators.scala
```diff
@@ -4,14 +4,33 @@ import java.net.{InetAddress, InetSocketAddress}
...
 trait ObjectGenerators {
 
+  object FullNodePeerFeature$ extends PeerFeature {
+    override val featureId: PeerFeature.Id = 1: Byte
+
+    override def serializer: Serializer[PeerFeature] = new Serializer[PeerFeature] {
+      override def toBytes(obj: PeerFeature): Array[Byte] = Array(1: Byte, 2: Byte, 3: Byte)
+
+      override def parseBytes(bytes: Array[Byte]): Try[PeerFeature] = Try {
+        assert(bytes(0) == 1)
```
catena2w: do not use assert here, as soon as it can be turned of in runtime.

src/main/scala/scorex/ObjectGenerators.scala
```diff
@@ -4,14 +4,33 @@ import java.net.{InetAddress, InetSocketAddress}
...
 
+import scala.util.Try
+
 trait ObjectGenerators {
 
+  object FullNodePeerFeature$ extends PeerFeature {
+    override val featureId: PeerFeature.Id = 1: Byte
+
+    override def serializer: Serializer[PeerFeature] = new Serializer[PeerFeature] {
+      override def toBytes(obj: PeerFeature): Array[Byte] = Array(1: Byte, 2: Byte, 3: Byte)
```
catena2w: (1,2,3) here looks like some magic numbers, why 1,2,3?

src/main/scala/scorex/ObjectGenerators.scala
```diff
@@ -4,14 +4,33 @@ import java.net.{InetAddress, InetSocketAddress}
...
 import scorex.core.transaction.box.proposition.PublicKey25519Proposition
 import scorex.core.transaction.state.{PrivateKey25519, PrivateKey25519Companion}
 import scorex.core.{ModifierId, ModifierTypeId, NodeViewModifier}
 import scorex.crypto.signatures.Curve25519
 
+import scala.util.Try
+
 trait ObjectGenerators {
 
+  object FullNodePeerFeature$ extends PeerFeature {
```
catena2w: Why `$` at the end?

src/main/scala/scorex/core/network/Handshake.scala
```diff
@@ -34,15 +31,26 @@ object HandshakeSerializer extends Serializer[Handshake] {
...
     val nodeNameBytes = obj.nodeName.getBytes
 
-    Array(anb.size.toByte) ++ anb ++
+    val featureBytes = obj.features.foldLeft(Array(obj.features.size.toByte)){case (fb, f) =>
+      val featId = f.featureId
+      val featBytes = f.bytes
+      fb ++ (featId +: Shorts.toByteArray(featBytes.length.toShort)) ++ featBytes
+    }
+
+    Array(anb.size.toByte) ++
```
catena2w: For performance reasons it's better to use `scorex.core.utils.concatBytes` or `com.google.common.primitives.Bytes.concat` instead of `++`

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -37,6 +38,10 @@ class NetworkController(settings: NetworkSettings,
 
   private implicit val system: ActorSystem = context.system
 
+  private val featureSerializers: PeerFeature.Serializers = features.map(f => f.featureId -> f.serializer).toMap
+  private val handshakeSerializer = new HandshakeSerializer(featureSerializers, settings.maxHandshakeSize)
+
+  //todo: make usage more clear
```
catena2w: I think you'd better write more comments here, because it's not so clear, what should be improved (and this will be forgotten in a future)

kushti: @catena2w I've addressed the issues reported, please make another pass!

kushti:
@catena2w I've addressed the issues reported, please make another pass!

PR #274, opened 2018-07-02
ModifiersCache: add value to onPut

catena2w:
value may be useful during `onPut` operation, if we're going to do something with other modifiers in a cache based on the inserted one. Example - prioritize modifiers that have inserted modifier as a parent.


catena2w:
Not so important

PR #276, opened 2018-07-04
Remove proposition

catena2w:


src/main/scala/scorex/core/app/Application.scala
```diff
@@ -20,7 +20,6 @@ trait Application extends ScorexLogging {
 
   import scorex.core.network.NetworkController.ReceivableMessages.ShutdownNetwork
 
-  type P <: Proposition
```
kushti: remove unused imports as well

catena2w: done

PR #277, opened 2018-07-05
Make network controller lazy to fix NPE

catena2w:


catena2w:
Can be fixed in application layer

PR #283, opened 2018-07-09
ToDownload processing

catena2w:


src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -295,8 +303,25 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
...
   }
 
+  /**
+    * Our node needs modifier of type `modifierTypeId` with id `modifierIds` but peer that can deliver
+    * it is unknown
+    */
+  protected def requestDownload(modifierTypeId: ModifierTypeId, modifierIds: Seq[ModifierId]): Unit = {
+    deliveryTracker.expect(modifierTypeId, modifierIds)
+    val msg = Message(requestModifierSpec, Right(modifierTypeId -> modifierIds), None)
+    //todo: Full nodes should be here, not a random peer
```
kushti: Scorex does not have a notion of a full-node, so the todo note should be like "A peer which is supposedly having the modifier should be here, not a random peer."

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -295,8 +303,25 @@ MR <: MempoolReader[TX]](networkControllerRef: ActorRef,
       }
   }
 
+  /**
+    * Our node needs modifier of type `modifierTypeId` with id `modifierIds` but peer that can deliver
```
kushti: needs for modifiers ... with ids

kushti:
@catena2w accepted, but please address the minor issues in the comments. Also, can we have a test for the new logic?

PR #289, opened 2018-07-13
ModifierId as String

### LINKED ISSUE #178: Wrapper for Array[Byte]

### LINKED ISSUE #208: Replace various forms of byte collections by `ByteString` and `Seq`.

catena2w:
Closes #208
Closes #178


examples/src/main/scala/examples/commons/SimpleBoxTransaction.scala
```diff
@@ -122,7 +123,7 @@ object SimpleBoxTransaction extends ScorexEncoding {
     val amount = to.map(_._2.toLong).sum
 
     val from: IndexedSeq[(PrivateKey25519, Nonce, Value)] = w.boxes()
-      .filter(b => !boxesIdsToExclude.exists(_ sameElements b.box.id)).sortBy(_.createdAt).takeWhile { b =>
+      .filter(b => !boxesIdsToExclude.contains(b.box.id)).sortBy(_.createdAt).takeWhile { b =>
```
mike-aksarin: if we need to be faster here, we could use Sets.

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -224,23 +234,25 @@ class NodeViewSynchronizer[TX <: Transaction,
       val modifiers = data._2
 
       log.info(s"Got modifiers of type $typeId from remote connected peer: $remote")
-      log.trace(s"Received modifier ids ${data._2.keySet.map(encoder.encode).mkString(",")}")
+      log.trace(s"Received modifier ids ${data._2.keySet.map(id => encoder.encode(id)).mkString(",")}")
```
mike-aksarin: Do we need `encoder.encode` here?

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -224,23 +234,25 @@ class NodeViewSynchronizer[TX <: Transaction,
...
+        statusKeeper.received(id)
+        deliveryTracker.onReceive(typeId, id, remote)
+      }
 
       val (spam, fm) = modifiers.partition { case (id, _) => deliveryTracker.isSpam(id) }
 
       if (spam.nonEmpty) {
         log.info(s"Spam attempt: peer $remote has sent a non-requested modifiers of type $typeId with ids" +
-          s": ${spam.keys.map(encoder.encode)}")
+          s": ${spam.keys.map(id => encoder.encode(id))}")
```
mike-aksarin: Do we need `encoder.encode` here?

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
```diff
@@ -31,16 +31,16 @@ import scala.util.Success
   * @tparam SIS SyncInfoMessage specification
   */
 class NodeViewSynchronizer[TX <: Transaction,
-                           SI <: SyncInfo,
```
mike-aksarin: is it automatic code formatting who aligns these lines that way?

src/main/scala/scorex/core/validation/ModifierValidator.scala
```diff
@@ -216,10 +217,16 @@ case class ValidationState[T](result: ValidationResult[T], strategy: ValidationS
...
     */
-  def demandEqualIds(given: => Array[Byte], expected: => Array[Byte], fatalError: String): ValidationState[T] = {
+  def demandEqualIds(given: => ModifierId, expected: => ModifierId, fatalError: String): ValidationState[T] = {
     validateEqualIds(given, expected)(d => ModifierValidator.fatal(fatalError, d))
   }
 
+  /** Shortcut `require`-like method to validate the arrays are equal. Otherwise returns fatal error
+    */
+  def demandEqualArrays(given: => Array[Byte], expected: => Array[Byte], fatalError: String): ValidationState[T] = {
+    validateEquals(given)(expected)(d => ModifierValidator.fatal(fatalError, d))
```
mike-aksarin: It's better to use `Arrays.equals` here while `validateEquals ` uses `sameElements`. It could look like:
```
validate(Arrays.equals(given, expected)){ ModifierValidator.fatal(s"$fatalError. Given $given while expected $expected") }
```

src/main/scala/scorex/core/validation/ModifierValidator.scala
```diff
@@ -94,11 +94,12 @@ object ModifierValidator extends ScorexLogging  {
   val success: Valid[Unit] = Valid(())
 
   private def msg(descr: String, e: Throwable): String = msg(descr, Option(e.getMessage).getOrElse(e.toString))
+
   private def msg(description: String, detail: String): String = s"$description: $detail"
 }
 
 /** This is the place where all the validation DSL lives */
-case class ValidationState[T](result: ValidationResult[T], strategy: ValidationStrategy)(implicit e: BytesEncoder) {
+case class ValidationState[T](result: ValidationResult[T], strategy: ValidationStrategy)(implicit e: ScorexEncoder) {
```
mike-aksarin: i'd like to drop `ScorexEncoder` here and use just `id.toString` method as `id` is not an array of bytes anymore

src/main/scala/scorex/core/validation/ModifierValidator.scala
```diff
@@ -26,7 +26,7 @@ import scala.util.Try
 trait ModifierValidator {
 
   /** Encoder from bytes to string and back for logging */
-  implicit val encoder: BytesEncoder
+  implicit val encoder: ScorexEncoder
```
mike-aksarin: i'd like to drop `encoder` here and use just `id.toString` method as `id` is not an array of bytes anymore

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
catena2w: I think it might be useful if we'll decide to switch from `Base16` in `ModifierId` to something else, or if desired enocoding will be different from `Base16`

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
catena2w: I think it might be useful if we'll decide to switch from Base16 in ModifierId to something else, or if desired enocoding will be different from Base16

examples/src/main/scala/examples/commons/SimpleBoxTransaction.scala
catena2w: It's ok for examples

src/main/scala/scorex/core/network/NodeViewSynchronizer.scala
catena2w: yes

catena2w:
replaced by https://github.com/ScorexFoundation/Scorex/pull/294/files

Issue #304, opened 2018-08-17
Change API errors content type to json 


mike-aksarin:
Currently API provide `application/json` content type for successful responses and default `text/plain` for error responses. We need to unify response content type.
1) Change response content-type in `ApiError.complete` method.
2) Override default akka http error handling.
3) Update example yaml file.

The same changes should be performed in Ergo after finishing this task.

PR #312, opened 2018-10-08
 Content type for error response changed.

oskin1:
#304

### DIFF
--- a/examples/src/main/resources/api/testApi.yaml
+++ b/examples/src/main/resources/api/testApi.yaml
@@ -58,6 +58,20 @@ components:
           items:
             type: string
             example: 127.0.0.1:27272
+    ApiError:
+      type: object
+      required:
+        - error
+        - reason
+      properties:
+        error:
+          type: int32
+          description: Error code
+          example: 500
+        reason:
+          type: string
+          description: Error reason description
+          example: 'Internal server error'
 
 
 paths:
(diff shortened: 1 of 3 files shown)


src/main/scala/scorex/core/api/http/ApiError.scala
```diff
@@ -12,12 +13,14 @@ case class ApiError(statusCode: StatusCode, reason: String = "") {
 
   def complete(detail: String = ""): Route = {
     val nonEmptyReason = if (reason.isEmpty) statusCode.reason else reason
-    val body = if (detail.isEmpty) nonEmptyReason else s"$nonEmptyReason $detail"
-    Directives.complete(statusCode.intValue() -> body)
+    val response = Map("error" -> statusCode.intValue.asJson, "reason" -> nonEmptyReason.asJson).asJson
```
mike-aksarin: we have lost `detail` field here

examples/src/main/resources/api/testApi.yaml
```diff
@@ -58,6 +58,20 @@ components:
           items:
             type: string
             example: 127.0.0.1:27272
+    ApiError:
+      type: object
+      required:
+        - error
+        - reason
```
mike-aksarin: I'd suggest adding optional `detail` field here. `Reason` would contain string error code, `detail` text description

PR #315, opened 2018-10-10
Custom API RejectionHandler.

oskin1:


oskin1:
#315

src/main/scala/scorex/core/api/http/ApiRejectionHandler.scala
```diff
@@ -0,0 +1,36 @@
+package scorex.core.api.http
+
+import akka.http.scaladsl.server._
+
+object ApiRejectionHandler {
+
+  implicit val rejectionHandler: RejectionHandler = RejectionHandler.newBuilder()
+    .handleAll[SchemeRejection] { rejections =>
+      val schemes = rejections.map(_.supported).mkString(", ")
+      ApiError.BadRequest.complete(s"Uri scheme not allowed, supported schemes: $schemes")
```
mike-aksarin: i suggest dropping `.complete` word to allow `apply` method call implicitly. Same for bellow

PR #322, opened 2018-11-25
Stop trying to connect to not responding peer

### LINKED ISSUE #301: Ignore not responding peers
If some peer is not online, but our node knows it(because we received it in Peers message or it is written in config), there are a lot of messages like:
```
Trying to clear status for <peer>, but it is not found
```

oskin1:
Close #301

### DIFF
--- a/scalastyle-config.xml
+++ b/scalastyle-config.xml
@@ -68,7 +68,7 @@
     </check>
     <check level="warning" class="org.scalastyle.scalariform.CyclomaticComplexityChecker" enabled="true">
         <parameters>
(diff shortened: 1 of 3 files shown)


oskin1:
In order to avoid repeated connections to not responding peers I propose to implement blacklist timeout and add such peers to blacklist for a 24h.
Then blacklist would look like:
`val blacklist = mutable.Map[String, (TimeProvider.Time, Timeout)]()`

catena2w:
Fixed in https://github.com/ScorexFoundation/Scorex/pull/329

PR #324, opened 2018-12-14
Introduced Node View actors

mike-aksarin:


src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
+  protected implicit val defaultAskTimeout: Timeout = Timeout(10.seconds)
+  protected implicit val executionContext: ExecutionContext = context.dispatcher
 
-  /**
-    * Restore a local view during a node startup. If no any stored view found
+  /** Restore a local history during a node startup. If no any stored view found
     * (e.g. if it is a first launch of a node) None is to be returned
     */
-  def restoreState(): Option[NodeView]
+  protected def restoreHistory(): Option[History]
```
oskin1: Shouldn't this method be moved in `StateHistoryActor` now, as there are no state and history variables now in NVH?

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
 
-  /**
-    * Hard-coded initial view all the honest nodes in a network are making progress from.
+  /** Restore a local state during a node startup. If no any stored view found
+    * (e.g. if it is a first launch of a node) None is to be returned
     */
-  protected def genesisState: NodeView
+  protected def restoreState(): Option[State]
```
oskin1: Same as for `restoreHistory `

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
+    */
+  protected def restoreVault(): Option[Vault]
 
-  protected def history(): HIS = nodeView._1
-
-  protected def minimalState(): MS = nodeView._2
+  /** Restore a local memory pool during a node startup. If no any stored view found
+    * (e.g. if it is a first launch of a node) None is to be returned
+    */
+  protected def restoreMempool(): Option[MPool]
```
oskin1: Same as for `restoreHistory `

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
 
-  protected def vault(): VL = nodeView._3
+  /** Hard-coded initial history all the honest nodes in a network are making progress from
+    */
+  protected def genesisHistory(): History
 
-  protected def memoryPool(): MP = nodeView._4
+  /** Hard-coded initial state all the honest nodes in a network are making progress from
+    */
+  protected def genesisState(): State
```
oskin1: How these direct readers are expected to be used from NVH when state and history aren't available directly from NVH?

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
-  protected def minimalState(): MS = nodeView._2
+  /** Restore a local memory pool during a node startup. If no any stored view found
+    * (e.g. if it is a first launch of a node) None is to be returned
+    */
+  protected def restoreMempool(): Option[MPool]
 
-  protected def vault(): VL = nodeView._3
+  /** Hard-coded initial history all the honest nodes in a network are making progress from
+    */
+  protected def genesisHistory(): History
```
oskin1: Same as for `genesisState`

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
-      case None =>
-        memoryPool().put(tx) match {
-          case Success(newPool) =>
-            log.debug(s"Unconfirmed transaction $tx added to the memory pool")
-            val newVault = vault().scanOffchain(tx)
-            updateNodeView(updatedVault = Some(newVault), updatedMempool = Some(newPool))
-            context.system.eventStream.publish(SuccessfulTransaction[TX](tx))
+  /** Hard-coded initial history all the honest nodes in a network are making progress from
+    */
+  protected def genesisMempool(): MPool
```
oskin1: Same as for `genesisState`

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -1,393 +1,306 @@
...
-  case class UpdateInformation(history: HIS,
-                               state: MS,
-                               failedMod: Option[PMOD],
-                               alternativeProgressInfo: Option[ProgressInfo[PMOD]],
-                               suffix: IndexedSeq[PMOD])
+  /** Underlying memory pool actor
+    */
+  protected val memoryPoolActor: ActorRef = createMemoryPoolActor()
+
+  protected var vault: Vault = restoreVault().getOrElse(genesisVault())
```
oskin1: Why isn't `vault` implemented as an actor?

src/main/scala/scorex/core/NodeViewHolder.scala
mike-aksarin: Now I think, you are right, and these should be moved to the correspondent actors

src/main/scala/scorex/core/NodeViewHolder.scala
mike-aksarin: This was expected to be used for component actors initialization, as in Hybrid example. But you're probably right, this should be moved to actors itself

src/main/scala/scorex/core/NodeViewHolder.scala
mike-aksarin: Yes, this definitely should be implemented as an actor, I just did not manage to do it in time. Please fix this

PR #329, opened 2019-02-11
Network protocol update

### LINKED ISSUE #301: Ignore not responding peers

catena2w:
Changes:
- Add version field to all p2p messages 
- Do not send messages which which version is higher then peer version
- Refactoring: extract common `PeerData` structure from `PeerInfo` and `Handshake`
- Refactoring: `HandshakeSerializer` => `HandshakeSpec` 
- Refactoring: make `InvData` and `ModifiersData` a case class instead of a tuple
- Distribute PeerData in Peers message
- Only send recently seen and connected peers on GetPeers request
- Remove not responding peers from local peers database (Close #301)
- Remove peer from local database on handshake parsing failure


src/main/scala/scorex/core/network/peer/PeerInfo.scala
```diff
@@ -2,19 +2,29 @@ package scorex.core.network.peer
...
+  lazy val handshake: Handshake = Handshake(peerData, lastSeen)
 
-  lazy val localAddress: Option[InetSocketAddress] = {
-    features.collectFirst { case LocalAddressPeerFeature(addr) => addr }
-  }
 }
+
+object PeerInfo {
+
+  def fromAddress(address: InetSocketAddress): PeerInfo = {
```
oskin1: It's not clear that this method should be used for peers from config only. It would be better to make it obvious.

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -79,46 +78,66 @@ class PeerManager(settings: ScorexSettings, scorexContext: ScorexContext) extend
...
-    case object RandomPeer
-    case class RandomPeers(hawMany: Int)
-    case class RandomPeerExcluding(excludedPeers: Seq[PeerInfo])
 
-    // apiInterface messages
-    case object GetAllPeers
+    /**
+      * Message to get peers from known peers map filtered by `choose` function
+      */
+    trait GetPeers[T] {
```
oskin1:
```suggestion
    sealed trait GetPeers[T] {
```

src/main/scala/scorex/core/network/peer/PeerInfo.scala
```diff
@@ -2,19 +2,29 @@ package scorex.core.network.peer
...
-                    connectionType: Option[ConnectionType] = None,
-                    features: Seq[PeerFeature] = Seq()) {
+/**
+  * Information about peer to be stored in PeerDatabase
+  *
+  * @param peerData       - general information about the peer
+  * @param lastSeen       - timestamp when this peer was last seen in the network
+  * @param connectionType - type of connection (Incoming/Outgoing) established to this peer if any
+  */
+case class PeerInfo(peerData: PeerData,
```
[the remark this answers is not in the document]
catena2w: `PeerInfo` is just stored information about peer, it is not necessary connected.

src/main/scala/scorex/core/network/peer/PeerManager.scala
catena2w: It might be useful to redefine it in concrete apps, e.g. we may want to create 'GetUTXOPeers' class in Ergo

src/main/scala/scorex/core/network/peer/PeerInfo.scala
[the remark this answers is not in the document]
oskin1: @catena2w, I mean, that it would be better to make it obvious that the difference between `PeerData` and `PeerInfo` is that in the latter the information about peer appears in the context of connection.

src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
```diff
@@ -99,16 +121,21 @@ object ModifiersSpec {
...
+  * The `Modifier` message is a reply to a `RequestModifier` message which requested these modifiers.
+  */
+class ModifiersSpec(maxMessageSize: Int) extends MessageSpecV1[ModifiersData] with ScorexLogging {
 
   import ModifiersSpec._
 
   override val messageCode: MessageCode = MessageCode
   override val messageName: String = MessageName
 
+
```
oskin1: Why?

src/main/scala/scorex/core/network/message/BasicMessagesRepo.scala
```diff
@@ -159,27 +193,69 @@ object GetPeersSpec extends MessageSpec[Unit] {
...
     (0 until length).map { _ =>
-      val address = InetAddress.getByAddress(r.getBytes(AddressLength))
-      val port = r.getUInt().toIntExact
-      new InetSocketAddress(address, port)
+      peerSpecSerializer.parse(r)
     }
   }
 }
+
+
```
oskin1: Redundant line

src/main/scala/scorex/core/network/peer/PeerInfo.scala
```diff
@@ -2,19 +2,33 @@ package scorex.core.network.peer
...
+  * @param connectionType - type of connection (Incoming/Outgoing) established to this peer if any
+  */
+case class PeerInfo(peerSpec: PeerSpec,
+                    lastSeen: Long,
+                    connectionType: Option[ConnectionType] = None) {
 
-  lazy val reachablePeer: Boolean = {
-    declaredAddress.isDefined || localAddress.isDefined
-  }
+  lazy val handshake: Handshake = Handshake(peerSpec, lastSeen)
```
oskin1: Not used

PR #340, opened 2019-05-17
P2P telemetry

oskin1:


kushti:
While events probing could be a good idea, it probably should be configurable.

PR #343, opened 2019-06-05
put all test files to single folder and delete it on exit

catena2w:
Clean files created during tests. Close https://github.com/ergoplatform/ergo/issues/707


testkit/src/main/scala/scorex/testkit/utils/FileUtils.scala
```diff
@@ -25,8 +35,25 @@ trait FileUtils {
     createTempDirForPrefix(prefix)
   }
 
+  /**
+    * Recurcively remove all files and directories in `root`
```
oskin1:
```suggestion
    * Recursively remove all files and directories in `root`
```

PR #358, opened 2020-01-26
Banning peers for sending adversarially constructed messages

kushti:
This PR is introducing MaliciousBehaviorException indicating malicious behavior of a peer. Such exception is thrown if a peer tries to send wrong magic bytes (so a peer is from another network), or trying to declare length of a following-up buffer to be negative, or providing incorrect checksum for a message. This PR also contains some fixes for penalizing logic.

### DIFF
--- a/src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
+++ b/src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
@@ -1,0 +1,5 @@
+package scorex.core.network
+
+class MaliciousBehaviourException {
+
+}
--- a/src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
+++ b/src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
@@ -1,5 +1,8 @@
 package scorex.core.network
 
-class MaliciousBehaviourException {
-
-}
+/**
+  * Custom exception to distinguish malicious behaviour of external peers from non-adversarial network issues
+  *
+  * @param message - exception message
+  */
+case class MaliciousBehaviorException(message: String) extends Exception(message)
--- a/src/main/scala/scorex/core/network/message/Message.scala
+++ b/src/main/scala/scorex/core/network/message/Message.scala
@@ -70,7 +70,7 @@ class MessageSerializer(specs: Seq[MessageSpec[_]], magicBytes: Array[Byte]) {
 
       //peer is from another network
       if (!java.util.Arrays.equals(magic, magicBytes)) {
-        throw MaliciousBehaviorException("Wrong magic bytes" + magic.mkString)
+        throw MaliciousBehaviorException(s"Wrong magic bytes, expected ${magicBytes.mkString}, got ${magic.mkString}")
       }
 
       //peer is trying to cause buffer overflow or breaking the parsing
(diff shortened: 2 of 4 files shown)


src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
```diff
@@ -0,0 +1,8 @@
+package scorex.core.network
+
+/**
+  * Custom exception to distinguish malicious behaviour of external peers from non-adversarial network issues
+  *
+  * @param message - exception message
```
aslesarenko: Doesn't sounds like valuable description. :) The description can contain some hints on what kind of information should be provided in the message. This will also help to understand where this exception should and shouldn't be used.

src/main/scala/scorex/core/network/peer/PeerManager.scala
```diff
@@ -152,7 +155,8 @@ object PeerManager {
                           blacklistedPeers: Seq[InetAddress],
                           sc: ScorexContext): Option[PeerInfo] = {
         val candidates = knownPeers.values.filterNot { p =>
-          excludedPeers.exists(_.peerSpec.address == p.peerSpec.address)
+          excludedPeers.exists(_.peerSpec.address == p.peerSpec.address) &&
+            blacklistedPeers.exists(addr => p.peerSpec.address.map(_.getAddress).contains(addr))
```
aslesarenko: Mind blowing naming, when trying to make sense out of it `... address.map(_.getAddress) ...`

src/main/scala/scorex/core/network/MaliciousBehaviorException.scala
kushti: There are no limits for the message semantics I guess.

src/main/scala/scorex/core/network/peer/PeerManager.scala
kushti: What do you propose?

aslesarenko: Renaming, but this is not related to this PR

src/main/scala/scorex/core/network/message/Message.scala
```diff
@@ -67,8 +67,16 @@ class MessageSerializer(specs: Seq[MessageSpec[_]], magicBytes: Array[Byte]) {
       val magic = it.getBytes(MagicLength)
       val msgCode = it.getByte
       val length = it.getInt
-      require(java.util.Arrays.equals(magic, magicBytes), "Wrong magic bytes" + magic.mkString)
-      require(length >= 0, "Data length is negative!")
+
+      //peer is from another network
+      if (!java.util.Arrays.equals(magic, magicBytes)) {
+        throw MaliciousBehaviorException("Wrong magic bytes" + magic.mkString)
```
greenhat:
```suggestion
        throw MaliciousBehaviorException(s"Wrong magic bytes, expected ${magicBytes.mkString}, got ${magic.mkString}")
```

src/main/scala/scorex/core/network/message/Message.scala
```diff
@@ -79,7 +87,11 @@ class MessageSerializer(specs: Seq[MessageSpec[_]], magicBytes: Array[Byte]) {
           val checksum = it.getBytes(ChecksumLength)
           val data = it.getBytes(length)
           val digest = Blake2b256.hash(data).take(ChecksumLength)
-          if (!java.util.Arrays.equals(checksum, digest)) throw new Error(s"Invalid data checksum length = $length")
+
+          //peer reported incorrect checksum
+          if (!java.util.Arrays.equals(checksum, digest)) {
+            throw MaliciousBehaviorException(s"Invalid data checksum length = $length")
```
greenhat: The error message seems off. Why `length` is being reported when `checksum != digest`?

kushti: Thanks, fixed!

PR #365, opened 2020-07-26
Fixing "too many open files" error in tests

kushti:
Currently "too many open files" error appears in tests. The error is rooting in how recursive deletion is implemented in scorex.testkit.utils.FileUtils.remove(). This PR fixes the problematic recursive files deletion. 

The fix is made by @knizhnik and extracted from the #362 .

### DIFF
--- a/testkit/src/main/scala/scorex/testkit/utils/FileUtils.scala
+++ b/testkit/src/main/scala/scorex/testkit/utils/FileUtils.scala
@@ -33,21 +33,21 @@ trait FileUtils {
     createTempDirForPrefix(prefix)
   }
 
+  /**
+    * Recursively remove all the files and directories in `root`
+    */
+  def remove(root: Path): Unit = {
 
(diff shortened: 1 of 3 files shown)


testkit/src/main/scala/scorex/testkit/utils/FileUtils.scala
```diff
@@ -35,21 +33,22 @@ trait FileUtils {
     createTempDirForPrefix(prefix)
   }
 
+
+  @SuppressWarnings(Array("org.wartremover.warts.Recursion"))
```
aslesarenko, line 37: Please add ScalaDoc, in particular describing the rationale behind not deleting some of the files.

kushti: Removed preserving hidden files (it was copy-pasted from the internets likely), also, made the deleteRecursive method internal to remove()

PR #369, opened 2020-09-13
Minor updates for networking

polarker:
Three minor updates

1. Enable pull model for incoming connections too. Right now, pull model is only enabled for outgoing connections.
2. Both confirmed and unconfirmed connections are considered when checking the number of connections.
3. Use `remoteAddress` as peer's address, if there is no address info in it's `PeerInfo`

### DIFF
--- a/src/main/scala/scorex/core/network/NetworkController.scala
+++ b/src/main/scala/scorex/core/network/NetworkController.scala
@@ -67,10 +67,11 @@ class NetworkController(settings: NetworkSettings,
   log.info(s"Declared address: ${scorexContext.externalNodeAddress}")
 
   //bind to listen incoming connections
-  tcpManager ! Bind(self, bindAddress, options = Nil, pullMode = false)
+  tcpManager ! Bind(self, bindAddress, options = Nil, pullMode = true)
 
-  override def receive: Receive =
-    bindingLogic orElse
+  override def receive: Receive = bindingLogic
+
+  def mainLogic(tcpListener: ActorRef): Receive =
       businessLogic orElse
       peerCommands orElse
       connectionEvents orElse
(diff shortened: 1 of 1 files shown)


src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -67,10 +67,11 @@ class NetworkController(settings: NetworkSettings,
   log.info(s"Declared address: ${scorexContext.externalNodeAddress}")
 
   //bind to listen incoming connections
-  tcpManager ! Bind(self, bindAddress, options = Nil, pullMode = false)
+  tcpManager ! Bind(self, bindAddress, options = Nil, pullMode = true)
```
scasplte2, line 70: Unless I am misunderstanding, I thought pull mode was purposely disabled here since to enable it would cause the network controller to respond to the TCP manager each time a new connection is attempted. But since each attempt to connect to peers is independent, there is no need to block subsequent connection requests.

Relevant documentation on pullMode in bind command 
https://doc.akka.io/docs/akka/2.6.9/io-tcp.html?language=java#pull-mode-reading-for-inbound-connections

`pullMode` on connect makes sense to me as we want the PeerConnectionHandlers between remote peers to fully process and acknowledge messages to maintain a stable connection.

polarker: Sorry for the late reply. Pull mode is supposed to throttle reading, not to throttle the number of connection requests. However, in `Akka`, both are throttled, so pull mode was disabled on purpose to not throttle the number of incoming connections.

In this pull request, pull mode is enabled to throttle only reading, and at the same time, sending a message to TCP manager to always accept new connections.

PR #370, opened 2020-09-19
Parse in synchronizers

scasplte2:
## Issue ##
Currently messages received from a remote peer are parsed by the NetworkController before being sent to the respective handlers. However, the message parsing does not need to happen at the NetworkController stage and can instead be passed to appropriate handler (aka a Synchronzier). In the current setup there are two issues that may be addressed:
1. At the NetworkController, only the message code contained in the message spec is used to identify the handler for a message. This message code is available when the PeerConnectionHandler forwards the message to NetworkController therefore NetworkController is parsing data needlessly.
1. Parsing an incoming message may be a time intensive procedure (particularly for large chunks of modifiers), thus stalling the NetworkController which may in turn delay other NetworkController responsibilities such as peer connections or handling egress traffic from our local node.

## Proposed solution ##
A Synchronizer trait has been created and implemented by NodeViewSynchronizer and PeerSychronizer. This trait defines a common definition for parsing a message received from a remote peer and handing the parsed data off to a specified function for processing. This also allows for us to get rid of the DataFromPeer class a use a single representation of message data for both ingress and egress traffic.


scasplte2:
Unsure why the remote Travis tests are failing (it appears to be a dependency issue) but local runs of 
```
sbt "project examples" clean coverage test && sbt test && sbt "project examples" coverageReport coveralls
```
appear to successfully pass. Perhaps a maintainer could trigger the run again to test if it is a temporary issue.

kushti:
> Unsure why the remote Travis tests are failing (it appears to be a dependency issue) but local runs of
> 
> ```
> sbt "project examples" clean coverage test && sbt test && sbt "project examples" coverageReport coveralls
> ```
> 
> appear to successfully pass. Perhaps a maintainer could trigger the run again to test if it is a temporary issue.

Travis failed to resolve a dependency, for unclear reasons.

kushti (approved):

kushti (approved):

Issue #379, opened 2020-10-25
Move max inactive connection deadline to settings


kushti:
Currently, *scheduleDroppingDeadConnections* in *NetworkController* has the following magic constant:

// A peer should send out sync message to us at least once per settings.syncStatusRefreshStable duration.
// We wait for more, namely settings.syncStatusRefreshStable.toMillis * 3
val timeout = settings.syncStatusRefreshStable.toMillis * 3

But better to have dedicated setting for detecting inactive connections, e.g. "inactiveConnectionDeadline"

kushti:
done

PR #380, opened 2020-10-26
[refer #379] Add inactiveConnectionDeadline configuration settings pa…

knizhnik:
…rameter


src/main/resources/reference.conf
```diff
@@ -144,6 +144,9 @@ scorex {
     # Synchronization status update interval for stable regime
     syncStatusRefreshStable = 4m
 
+    # Dropping dead connections timeout
```
kushti, line 147: Timeout for dropping dead connections

src/main/scala/scorex/core/api/http/PeersApiRoute.scala
```diff
@@ -36,14 +37,19 @@ case class PeersApiRoute(peerManager: ActorRef,
...
+        con.peerInfo match {
+          case Some(peerInfo) =>
+            PeerInfoResponse(
+              address = peerInfo.peerSpec.declaredAddress.map(_.toString).getOrElse(""),
+              lastMessage = con.lastMessage,
+              lastHandshake = peerInfo.lastHandshake,
+              name = peerInfo.peerSpec.nodeName,
+              connectionType = peerInfo.connectionType.map(_.toString)
+            )
+          case None => throw new Exception("Peer is not connected")
```
kushti, line 51: We shouldn't get here, but throwing exceptions in API is not good, log.error is enough (thus connected peers will be returned)

src/main/scala/scorex/core/network/peer/PeerInfo.scala
```diff
@@ -9,11 +9,11 @@ import scorex.core.network.{ConnectionDirection, PeerSpec}
   * Information about peer to be stored in PeerDatabase
   *
   * @param peerSpec       - general information about the peer
-  * @param lastSeen       - timestamp when this peer was last seen in the network
+  * @param lastHandshake       - timestamp when this peer was last seen in the network
```
kushti, line 12: timestamp when last handshake was done (last seen sounds ambiguous, i.e. can be considered as time when we got a message from a peer last time)

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -242,14 +238,14 @@ class NetworkController(settings: NetworkSettings,
...
-        val lastSeen = cp.peerInfo.map(_.lastSeen).getOrElse(now)
-        // A peer should send out sync message to us at least once per settings.syncStatusRefreshStable duration.
-        // We wait for more, namely settings.syncStatusRefreshStable.toMillis * 3
-        val timeout = settings.syncStatusRefreshStable.toMillis * 3
-        val delta = now - lastSeen
-        if (delta > timeout) {
-          log.info(s"Dropping connection with ${cp.peerInfo}, last seen ${delta / 1000.0} seconds ago")
-          cp.handlerRef ! CloseConnection
+        val lastSeen = cp.lastMessage
+        if (lastSeen != 0) {
```
kushti, line 242: Then if no messages ever received from a peer and connection is dead, the connection will never be removed. I suggested to set lastMessage = now then creating connection, and then the check is not needed.

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -316,7 +312,7 @@ class NetworkController(settings: NetworkSettings,
 
     val handler = context.actorOf(handlerProps) // launch connection handler
     context.watch(handler)
-    val connectedPeer = ConnectedPeer(connectionId, handler, None)
+    val connectedPeer = ConnectedPeer(connectionId, handler, 0, None)
```
kushti, line 315: Why None? Maybe possible to get peer info from PeerManager (ConfirmConnection handling)

kushti (changes requested):

src/main/scala/scorex/core/api/http/PeersApiRoute.scala
knizhnik: Sorry, I do not know how to replace throw with just logging (because otherwise  compiler reports error).
Actually I need to call "get", because Option is this case is guaranteed to be non empty. But using Option.get is now prohibited. May be you can suggest some other better to handle this situation in Scala (transform components of a sequence, ignoring elements which Option fields are empty).

src/main/resources/reference.conf
knizhnik: Done

src/main/scala/scorex/core/network/peer/PeerInfo.scala
knizhnik: Done

src/main/scala/scorex/core/network/NetworkController.scala
knizhnik: Done

src/main/scala/scorex/core/network/NetworkController.scala
knizhnik: peerInfo will be set when handshake is established.
peerInfo.notEmpty is used now to filter connected peers.
I wonder if we need to distinguish connected and handshaked peers?

src/main/scala/scorex/core/api/http/PeersApiRoute.scala
kushti: ah I see, replaced .map with flatmap, and not checking errors at all (such checks should be done in other places)

kushti (approved):

Issue #382, opened 2020-11-02
Introducing session id and network id during the handshake


kushti:
A handshake can be enhanced via adding a new PeerFeature, see LocalAddressPeerFeature. Do a new feature with id = 3 which contains network magic bytes (from settings) and a session ID. The session ID should be generated when the node started, 64 random bits (8 bytes) enough. If session ID is not received from other peer, then it equals to 0. 

Then the following code needs to be rewritten, isSelf should be based on session Id, not address. 

 //drop connection to self if occurred or peer already connected
      val shouldDrop = isSelf(remoteAddress) ||
        connectionForPeerAddress(peerAddress).exists(_.handlerRef != peerHandlerRef)
      if (shouldDrop) {
        connectedPeer.handlerRef ! CloseConnection
        peerManagerRef ! RemovePeer(peerAddress)
        connections -= connectedPeer.connectionId.remoteAddress
      } else {
        peerManagerRef ! AddOrUpdatePeer(peerInfo)

        val updatedConnectedPeer = connectedPeer.copy(peerInfo = Some(peerInfo))
        connections += remoteAddress -> updatedConnectedPeer
        context.system.eventStream.publish(HandshakedPeer(updatedConnectedPeer))
      }

Also, connection should be dropped if handshake is coming from another network (different magic)

Also for the code above, if peer already connected, "
        peerManagerRef ! RemovePeer(peerAddress)
        connections -= connectedPeer.connectionId.remoteAddress" is not needed (it is needed for self only)

PR #384, opened 2020-11-10
[refer #382] Add SessionIdPeerFeature

knizhnik:


src/main/scala/scorex/core/network/peer/SessionIdPeerFeature.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object SessionIdPeerFeatureSerializer extends ScorexSerializer[SessionIdPeerFeature] {
+  override def serialize(obj: SessionIdPeerFeature, w: Writer): Unit = {
+    w.putInt(obj.networkMagic.size)
+    w.putBytes(obj.networkMagic)
+    w.putLong(obj.sessionId)
+  }
+
+  override def parse(r: Reader): SessionIdPeerFeature = {
+    val networkMagic = r.getBytes(r.getInt())
```
kushti, line 35: better to use putUInt / getUInt

src/main/scala/scorex/core/network/NetworkController.scala
```diff
@@ -290,13 +291,11 @@ class NetworkController(settings: NetworkSettings,
           s"New outgoing connection to ${connectionId.remoteAddress} established (bound to local ${connectionId.localAddress})"
       }
     }
-
     val isLocal = connectionId.remoteAddress.getAddress.isSiteLocalAddress ||
       connectionId.remoteAddress.getAddress.isLoopbackAddress
-    val peerFeatures =
-      if (isLocal) scorexContext.features :+ LocalAddressPeerFeature(
+    var peerFeatures = scorexContext.features :+ mySessionIdFeature
```
kushti, line 296: peerFeatures can be immutable (val) still

src/main/scala/scorex/core/network/peer/SessionIdPeerFeature.scala
```diff
@@ -0,0 +1,39 @@
...
+    val sessionId = r.getLong()
+    SessionIdPeerFeature(networkMagic, sessionId)
+  }
+}
```
kushti, line 39: Please do a test for serialization (round-trip at least)

kushti (changes requested):

src/main/scala/scorex/core/network/peer/SessionIdPeerFeature.scala
knizhnik: getUInt() returns Long and can not be passed to getBytes

kushti: Actually, network magic is always about 4 bytes, so no need to pass this value

kushti (approved):

Issue #393, opened 2020-12-17
Peer eviction 


kushti:
To randomize connection, we need to evict a random peer periodically (e.g. every few hours) like it is done in Bitcoin

PR #394, opened 2020-12-18
[refer #393] Support periodic random connection eviction

knizhnik:


kushti (changes requested):
Looks good in general, but I need to test it (also, consider writing tests for the code). Please do PR description (and correct title maybe , as PR is not just about eviction). Also, please do comments, left some notes on that.

PR #397, opened 2021-07-30
handle db errors at higher level

pragmaxim:
@kushti This is the most important part to review. `reportModifierIsValid` and `reportModifierIsInvalid` now return `Try` so this is the "higher level" place where we propagated the DB errors to. I did not publish any event as this is neither `semantically` nor `syntactically` failed modification, it just cannot be persisted. 

It differs as now DB errors leads to returning : 
```
                UpdateInformation(history, updateInfo.state, Some(modToApply), None, updateInfo.suffix)
```
which would lead to another `updateState` call which is probably wrong, it is just to demonstrate that we are handling that effect now.

### DIFF
--- a/src/main/scala/scorex/core/NodeViewHolder.scala
+++ b/src/main/scala/scorex/core/NodeViewHolder.scala
@@ -248,13 +248,23 @@ trait NodeViewHolder[TX <: Transaction, PMOD <: PersistentNodeViewModifier]
       if (updateInfo.failedMod.isEmpty) {
         updateInfo.state.applyModifier(modToApply) match {
           case Success(stateAfterApply) =>
-            val newHis = history.reportModifierIsValid(modToApply)
-            context.system.eventStream.publish(SemanticallySuccessfulModifier(modToApply))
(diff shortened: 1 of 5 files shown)


kushti (approved):

pragmaxim:
I reviewed it one more time and there was a bug ... fixed in the latest commit

src/main/scala/scorex/core/NodeViewHolder.scala
```diff
@@ -294,8 +304,8 @@ trait NodeViewHolder[TX <: Transaction, PMOD <: PersistentNodeViewModifier]
 
               case Failure(e) =>
                 log.warn(s"Can`t apply persistent modifier (id: ${pmod.encodedId}, contents: $pmod) to minimal state", e)
+                // not publishing SemanticallyFailedModification as this is an internal error
                 updateNodeView(updatedHistory = Some(newHistory))
-                context.system.eventStream.publish(SemanticallyFailedModification(pmod, e))
```
pragmaxim, line 298: @kushti I removed this because this Failure can only happen due to some unexpected internal error, we already fire it in the `applyState` method

